Issue #40, opened 2016-05-27
Pubmed search


pkraker:
It would be awesome to extend the search functionalty to Pubmed. This would require to conduct a search in Pubmed and to retrieve basic metadata, abstract and categories for the first 100 papers. For display purposes in the visualization, we will also need to retrieve the full text PDF on demand (if it's available).

sckott:
Sounds good. how best to integrate into the codebase?

pkraker:
I have now written a modular version of the R scripts. The idea is to have a script for every API. Each script provides a get_papers function that returns metadata and text for the search results.

I have created a stub for the Pubmed search that includes a description: https://github.com/pkraker/Headstart/blob/master/server/preprocessing/other-scripts/pubmed.R

For reference, see the PLOS script that implements the get_papers() function: https://github.com/pkraker/Headstart/blob/master/server/preprocessing/other-scripts/rplos_fast.R

For testing purposes, have a look at the following script. It performs the whole query and analysis part: https://github.com/pkraker/Headstart/blob/master/server/preprocessing/other-scripts/test/rplos_fast-test.R

sckott:
working on it

pkraker:
Nice! I had to switch out xml2::xml_find_first with xml2::xml_find_one to get it to work. Which version of xml2 are you using? Also, searching with query="tomato" and params=NULL leads to a warning (and a subsequent error) that there are rows with the same name.

sckott:
Sorry, i am on the dev version of `xml2`. and that fxn name changed in the dev version.

> Also, searching with query="tomato" and params=NULL leads to a warning (and a subsequent error) that there are rows with the same name.

did you already update the function, so i can test what you have

pkraker:
Sorry, just pushed my changes. Maybe some of my other edits are to blame.

sckott:
using the latest code you have that works for me, let me know if that still is broken for you

pkraker:
Thanks, let me close this one then. I'll open separate issues, if I run into any problems.

PR #49, opened 2016-06-07
#40 added code for pubmed search

sckott:
@pkraker let me know what you think. 

* url is to the NCBI landing page for the article using the PMID - do you want the landing page to be the publisher's page? if so, they don't give that, but we could just use doi.org + doi 
* i added `date` and `year` fields since you had a field `year`, but I think you had a full date in there, so i added sep. fields for full date and just year
* it's a bit messier than PLOS as we have to parse XML, but it's i think still somewhat fast


pkraker:
Awesome, I am going to take it for a spin :)

PR #84, opened 2017-01-10
first version of evaluation pipeline

chreman:
consists of
* script to download baseline sampleset
* script to produce baseline metrics


chreman:
added a small tryCatch to enable both Rstudio and Rscript to execute the scripts

PR #85, opened 2017-01-11
Integrate io class

rbachleitner:
Moves data transformation logic and an asynchronous call to a new io class
to further separation of concerns.
One or two simplifications in headstart.js


rbachleitner:
Alright, I think I disregarded TimeLineView altogether, will check!

rbachleitner:
TimeLineView should work now as before.
List import was removed from io.js imports.

PR #92, opened 2017-01-16
Canvas class

rbachleitner:
Move chart initializing and drawing methods from headstart to canvas class.
Move event listener init from headstart to canvas class.
Change headstart method references in other module to canvas references.


rbachleitner:
@pkraker Info popup issue should be resolved now

rbachleitner:
@pkraker I think i fixed server_files mode can you check please?

PR #96, opened 2017-01-30
Refactoring a lot for mediator

rbachleitner:
- Delegates many calls from papers/bubbles/list to mediator;
- Adds render_list and render_bubbles option to default-config.js
- Mediator now only calls papers/bubbles/list if configured.
- Moves config from headstart to config
- headstart.js was simplified


pkraker:
It's looking very good, thanks! Please provide an example for the list-only mode.

rbachleitner:
@pkraker The last two commits should fix the things we talked about

rbachleitner:
Except for getting back from an enlarged paper in list only mode.

Bubblbu:
Awesome work! Looks great :+1: 

I have seen a few leftover cross-module calls (papers and canvas imported in bubbles.js, canvas.js and papers.js). Is it because of problems, when moved to mediator or are they left in the modules on purpose?

rbachleitner:
Thank you. To be honest it's because the separation of the modules is not finished yet. canvas, bubbles and papers are still very much intertwined and it'll be quite some work to disentangle them. My first goal however was to get the list out of there, the rest will follow.

pkraker:
We've come a long way with this PR, but there are still a few issues that need fixing:

**Major:**
* When I try to open the preview in the `local_files` example (paper "Personal Learning Environments - the future of eLearning ?"), I get the text for "What this?"
* The `search_repos` example does not work any more. The visualization is not shown. Most likely `headstart.php` is missing `headstart.start()`
* In the Pubmed example, all the papers are suddenly open access. This could be due to a change in Pubmed, but on the Open Knowledge Maps website, it still works as expected.
* When I try to preview papers in the Pubmed example, all of them are unavailable.
* When I view a single paper in the Pubmed example and zoom out again, the preview image is not hidden in the list.

**Minor:**
* Please resolve the conflict in `webpack.config.js`
* `debug` should be set to `false` in the `default-config.js`
* Please restore the `return d.y * -1;` on line 232 in `io.js`. This enables the standard look of the `local_files` examples.

rbachleitner:
@pkraker  commits ecfb5f7 - 121b9ee should fix everything you mention.

rbachleitner:
@pkraker Commits ecfb5f7 to b0eabeb should fix everything you mentioned.

pkraker:
Looking good!

PR #100, opened 2017-02-16
R experiments

chreman:
* merge master into R_experiments
* additional R scripts for evaluation/statistics of raw data, not very sophisticated yet but added for reproducibility


pkraker:
Nice!

Issue #28, opened 2016-04-27
Jumping between a single paper and all papers in the bubble triggers exception


pkraker:
Exception: "uncaught exception: event zoom inappropriate in current state enlarged"

Steps to reproduce:
1. Click on bubble
2. Click on a paper title in the list
3. Enter search term
4. Click on the bubble
5. Click on the same paper in the list
6. Click on the bubble again

pkraker:
The exception is gone, but the visualization now zooms out after step 4, which it shouldn't do. It should rather display the filtered list.

Bubblbu:
Interestingly PR #107 changed this insofar, that *sometimes* step 4 leads to the expected behaviour and sometimes still fails :confused:

chreman:
@pkraker can this issue be closed? I could not reproduce the bug with the given steps on a four maps (machine learning & digital education on base & pubmed)

pkraker:
@chreman Thanks for the pointer, I could also not reproduce this bug any more.

PR #107, opened 2017-02-20
Few fixes for minor issues

### LINKED ISSUE #98: List: filtering by year fails on last digit
As originally reported by @schmelling in OpenKnowledgeMaps/project-website#17:

"Hi this is only a small reminder.

We just tried out open knowledge maps at opencon berlin and think we found a small bug.

After the literatur search we wanted to filter by year through the feature tabs on the right and it returned 0 papers fast typing '2016' or '2015'. However '201' worked for showing the last decade.

Filtering by options in the beginning works fine.

Thanks for the great tool.
Cheers,
Nic"

### LINKED ISSUE #99: local_files: filtering has unexpected results
When a paper is selected in the local_files example, and the user enters a search term in the box, all the papers from this bubble matching the search term are shown in the list. The correct behaviour would be to show just the enlarged paper. This is also the way it works in the search_repos example. May be related to #28.

Bubblbu:
+ Fixes #99 (filtering of list while paper selected)
+ Fixes #98 (filtering by year)
+ Some mysterious behaviour for #28. See issue thread

### DIFF
--- a/vis/js/mediator.js
+++ b/vis/js/mediator.js
@@ -405,7 +405,6 @@ MyMediator.prototype = {
         mediator.current_bubble.current = "hoverbig";
         mediator.manager.call('papers', 'mouseoverpaper', []);
         mediator.manager.call('list', 'enlargeListItem', [d]);
-        d.paper_selected = true;
         mediator.current_enlarged_paper = d;
         mediator.manager.call('papers', 'framePaper', [d]);
     }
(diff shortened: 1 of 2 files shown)


vis/js/mediator.js
```diff
@@ -405,6 +405,7 @@ MyMediator.prototype = {
         mediator.current_bubble.current = "hoverbig";
         mediator.manager.call('papers', 'mouseoverpaper', []);
         mediator.manager.call('list', 'enlargeListItem', [d]);
+        d.paper_selected = true;
         mediator.current_enlarged_paper = d;
```
rbachleitner: This should probably go to `list.enlargeListItem` in the end.

Bubblbu: Agree! :+1:

pkraker:
Nice work!

PR #108, opened 2017-02-21
Add initializeMissingData method; Add default values to default-confi…

rbachleitner:
…g.js; Small fix in list.drawTitle;


rbachleitner:
@pkraker The default values are up for debate.

pkraker:
Looking good!

PR #110, opened 2017-02-21
Clusternaming

chreman:
* reworked missing keyword routine: now cleaner and returning better keywords
* added TF-IDF cluster naming based on titles, keywords, and their bi/trigrams
* added cascading filter from longer to shorter ngrams
* formatting gets lost, so included simple first-letter uppercasing
* some old code removed
* tested only locally, result for "natural language processing" on pubmed attached
[natural language processing_pubmed_ntn_clustering.pdf](https://github.com/OpenKnowledgeMaps/Headstart/files/791844/natural.language.processing_pubmed_ntn_clustering.pdf)

### DIFF
--- a/server/preprocessing/other-scripts/vis_layout.R
+++ b/server/preprocessing/other-scripts/vis_layout.R
@@ -216,7 +216,13 @@ create_cluster_labels <- function(clusters, metadata_full_subjects, weightingspe
   tfidf_top_names <- lapply(tfidf_top_names, function(x) {x = gsub("_", " ", x); trim(x)})
   tfidf_top_names <- lapply(tfidf_top_names, function(x) filter_out_nested_ngrams(x, top_n))
   tfidf_top_names <- lapply(tfidf_top_names, function(x) {paste0(toupper(substr(x, 1, 1)), substr(x, 2, nchar(x)))})
-  clusters$cluster_labels <- lapply(tfidf_top_names, function(x) {paste(unlist(trim(x)), collapse=", ")})
+  tfidf_top_names <- lapply(tfidf_top_names, function(x) {paste(unlist(trim(x)), collapse=", ")})
+  clusters$cluster_labels = ""
+  for (k in seq(1, clusters$num_clusters)) {
+    group = c(names(clusters$groups[clusters$groups == k]))
+    matches = which(clusters$labels%in%group)
+    clusters$cluster_labels[c(matches)] = tfidf_top_names[k]
+  }
   return(clusters)
 }
 
@@ -253,7 +259,7 @@ create_output <- function(clusters, layout, metadata) {
   result = cbind(x,y,groups,labels)
   output = merge(metadata, result, by.x="id", by.y="labels", all=TRUE)
   names(output)[names(output)=="groups"] <- "area_uri"
-  output["area"] = paste("Cluster ", output$area_uri, sep="")
+  output["area"] = cluster_labels
 
   output_json = toJSON(output)
 
--- a/server/preprocessing/other-scripts/vis_layout.R
+++ b/server/preprocessing/other-scripts/vis_layout.R
@@ -64,10 +64,6 @@ create_tdm_matrix <- function(metadata, text, sparsity=1, lang="english") {
   return(list(tdm_matrix = tdm_matrix, metadata_full_subjects = metadata_full_subjects))
 }
 
-BigramTokenizer <- function(x) unlist(lapply(ngrams(words(x), 2), paste, collapse = " "), use.names = FALSE)
-
-TrigramTokenizer <- function(x) unlist(lapply(ngrams(words(x), 3), paste, collapse = " "), use.names = FALSE)
-
 replace_keywords_if_empty <- function(corpus, metadata) {
 
   missing_subjects = which(lapply(metadata$subject, function(x) {nchar(x)}) <= 1)
--- a/server/preprocessing/other-scripts/vis_layout.R
+++ b/server/preprocessing/other-scripts/vis_layout.R
@@ -19,48 +19,48 @@ vis_layout <- function(text, metadata, max_clusters=15, maxit=500, mindim=2, max
   result <- create_tdm_matrix(metadata, text);
   metadata_full_subjects = result$metadata_full_subjects
   metadata_full_subjects_tfidf_top5 <- create_node_names(metadata_full_subjects)
-  
+
   print("normalize matrix")
   normalized_matrix <- normalize_matrix(result$tdm_matrix);
-  
+
   print("create clusters")
   clusters <- create_clusters(normalized_matrix, max_clusters=15);
   layout <- create_ordination(normalized_matrix, maxit=500, mindim=2, maxdim=2)
   output <- create_output(clusters, layout, metadata_full_subjects)
-  
+
   return(output)
-  
+
 }
 
 create_tdm_matrix <- function(metadata, text, sparsity=1, lang="english") {
   m <- list(content = "content", id = "id")
-  
+
   myReader <- readTabular(mapping = m)
-  
+
   (corpus <- Corpus(DataframeSource(text), readerControl = list(reader = myReader)))
-  
+
   corpus <- tm_map(corpus, removePunctuation)
-  
+
   corpus <- tm_map(corpus, removeWords, stopwords(lang))
-  
+
   corpus <- tm_map(corpus, stripWhitespace)
-  
+
   metadata_full_subjects <- replace_keywords_if_empty(corpus, metadata)
-  
+
   # corpus <- tm_map(corpus, content_transformer(tolower))
-  
+
   corpus_unstemmed = corpus
-  
+
   corpus <- tm_map(corpus, stemDocument)
-  
+
   tdm <- TermDocumentMatrix(corpus)
-  
+
   if(sparsity < 1) {
     tdm <- removeSparseTerms(tdm, sparsity)
   }
-  
+
   tdm_matrix = t(as.matrix(tdm))
-  
+
   return(list(tdm_matrix = tdm_matrix, metadata_full_subjects = metadata_full_subjects))
 }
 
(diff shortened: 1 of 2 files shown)


server/preprocessing/other-scripts/vis_layout.R
```diff
@@ -18,130 +18,189 @@ vis_layout <- function(text, metadata, max_clusters=15, maxit=500, mindim=2, max
...
+    # for ngrams: we have to collapse with "_" or else tokenizers will split ngrams again at that point and we'll be left with unigrams
+    texts_bigrams = lapply(lapply(texts, function(x)unlist(lapply(ngrams(unlist(strsplit(x, split=" ")), 2), paste, collapse="_"))), paste, collapse=" ")
+    texts_trigrams = lapply(lapply(texts, function(x)unlist(lapply(ngrams(unlist(strsplit(x, split=" ")), 3), paste, collapse="_"))), paste, collapse=" ")
+
+    subjects = paste(metadata_full_subjects$subject[c(matches)], texts_bigrams, texts_trigrams, collapse=" ")
+    subjects = gsub(",", ";", subjects)
+    subjectlist = c(subjectlist, subjects)
+  }
+  nn_corpus <- Corpus(VectorSource(subjectlist))
+  nn_corpus <- tm_map(nn_corpus, removeWords, stopwords("english"))
```
pkraker: Stopwords should have been filtered out at this point.

server/preprocessing/other-scripts/vis_layout.R
```diff
@@ -18,130 +18,189 @@ vis_layout <- function(text, metadata, max_clusters=15, maxit=500, mindim=2, max
...
 }
 
+
+SplitTokenizer <- function(x) {
+  tokens = unlist(lapply(strsplit(words(x), split=";"), paste), use.names = FALSE)
+  return(tokens)
+}
+
+
+create_cluster_names <- function(clusters, metadata_full_subjects, weightingspec, top_n) {
```
pkraker: This function should have an additional parameter: taxonomy_separator. This is necessary for sources like PLOS that report the full taxonomy tree with each keyword, e.g. Animals/Birds/Songbirds. In the function, keywords should be reduced to the last keyword, in our case Songbirds. See `KeywordNaming.php` lines 57-59.

server/preprocessing/other-scripts/vis_layout.R
```diff
@@ -18,130 +18,189 @@ vis_layout <- function(text, metadata, max_clusters=15, maxit=500, mindim=2, max
...
+  for (k in seq(1, clusters$num_clusters)) {
+    group = c(names(clusters$groups[clusters$groups == k]))
+    matches = which(metadata_full_subjects$id%in%group)
+
+    titles =  metadata_full_subjects$title[c(matches)]
+    subjects = metadata_full_subjects$subject[c(matches)]
+    texts = mapply(paste, titles, subjects)
+    texts = lapply(texts, function(x) {gsub("[^[:alnum:]]", " ", x)})
+
+    texts = lapply(texts, function(x)paste(removeWords(x, stopwords('english')), collapse=""))
```
pkraker: Stopwords should have been filtered out at this point.

server/preprocessing/other-scripts/vis_layout.R
```diff
@@ -18,130 +18,189 @@ vis_layout <- function(text, metadata, max_clusters=15, maxit=500, mindim=2, max
...
+  candidates = lapply(candidates, function(x)paste(unlist(strsplit(x, split="  ")), collapse=" "))
+  candidates_bigrams = lapply(lapply(candidates, function(x)unlist(lapply(ngrams(unlist(strsplit(x, split=" ")), 2), paste, collapse="_"))), paste, collapse=" ")
+  candidates = mapply(paste, candidates, candidates_bigrams)
+  candidates = lapply(candidates, function(x) {gsub('\\b\\d+\\s','', x)})
+
+  nn_corpus <- Corpus(VectorSource(candidates))
+  nn_tfidf <- TermDocumentMatrix(nn_corpus, control = list(tokenize = SplitTokenizer, weighting = function(x) weightSMART(x, spec="ntn")))
+  tfidf_top <- apply(nn_tfidf, 2, function(x) {x2 <- sort(x, TRUE);x2[x2>=x2[3]]})
+  tfidf_top_names <- lapply(tfidf_top, names)
+  replacement_keywords <- lapply(tfidf_top_names, FUN = function(x) {paste(unlist(x), collapse="; ")})
```
pkraker: The space after `;` in `collapse="; "` is superflouos, I think.

server/preprocessing/other-scripts/vis_layout.R
```diff
@@ -18,130 +18,189 @@ vis_layout <- function(text, metadata, max_clusters=15, maxit=500, mindim=2, max
...
+
   corpus <- tm_map(corpus, removeWords, stopwords(lang))
-  
-  metadata_full_subjects <- replace_keywords_if_empty(corpus, metadata)
-  
+
   corpus <- tm_map(corpus, stripWhitespace)
-  
+
+  metadata_full_subjects <- replace_keywords_if_empty(corpus, metadata)
```
chreman: will re-insert the old function as out-commented

server/preprocessing/other-scripts/vis_layout.R
```diff
@@ -18,130 +18,189 @@ vis_layout <- function(text, metadata, max_clusters=15, maxit=500, mindim=2, max
...
+
+  # corpus <- tm_map(corpus, content_transformer(tolower))
```
chreman: will be moved back

server/preprocessing/other-scripts/vis_layout.R
```diff
@@ -18,130 +18,189 @@ vis_layout <- function(text, metadata, max_clusters=15, maxit=500, mindim=2, max
...
-  
+
   return(list(tdm_matrix = tdm_matrix, metadata_full_subjects = metadata_full_subjects))
 }
 
+replace_keywords_if_empty <- function(corpus, metadata) {
+
+  missing_subjects = which(lapply(metadata$subject, function(x) {nchar(x)}) <= 1)
 
+  candidates = mapply(paste, metadata$title[missing_subjects], metadata$paper_abstract[missing_subjects])
```
chreman: content from pubmed will contain `df$content <- paste(df$title, df$paper_abstract, df$authors, df$subject, df$published_in, sep= " ")`
content from doaj will contain `text$content = paste(metadata$title, metadata$paper_abstract, metadata$published_in, metadata$authors, metadata$subject, sep=" ")`
which is marginally different but possibly a bit more noisy. will have a quick check on differences

server/preprocessing/other-scripts/vis_layout.R
```diff
@@ -18,130 +18,189 @@ vis_layout <- function(text, metadata, max_clusters=15, maxit=500, mindim=2, max
   print("calc matrix")
   result <- create_tdm_matrix(metadata, text);
   metadata_full_subjects = result$metadata_full_subjects
-  
+  metadata_full_subjects_tfidf_top5 <- create_node_names(metadata_full_subjects)
```
pkraker: The function `create_node_names` is not defined. Did you mean `create_cluster_names`?

server/preprocessing/other-scripts/vis_layout.R
```diff
@@ -18,130 +18,189 @@ vis_layout <- function(text, metadata, max_clusters=15, maxit=500, mindim=2, max
...
+
+create_cluster_names <- function(clusters, metadata_full_subjects, weightingspec, top_n) {
+  subjectlist = list()
+  for (k in seq(1, clusters$num_clusters)) {
+    group = c(names(clusters$groups[clusters$groups == k]))
+    matches = which(metadata_full_subjects$id%in%group)
+
+    titles =  metadata_full_subjects$title[c(matches)]
+    subjects = metadata_full_subjects$subject[c(matches)]
+    texts = mapply(paste, titles, subjects)
```
pkraker: Why do you add the titles to keywords at this point?

server/preprocessing/other-scripts/vis_layout.R
chreman: adding titles introduced an additional source of stopwords

server/preprocessing/other-scripts/vis_layout.R
chreman: This was based on our discussion with @brembs that keywords may not be totally representative of content, as they could be used similar to search engine optimization. I added paper titles as a source of keywords more accurately describing the content to be expected.

server/preprocessing/other-scripts/vis_layout.R
chreman: yes. this was a residual from my local testing environment. using tests from `test` flagged this up later. renamed functions and variables

server/preprocessing/other-scripts/vis_layout.R
chreman: included a taxonomy separator function, which filters out possible taxons based on the provided taxonomy separator, keeps the last branch and merge it back to the rest of the keywords

server/preprocessing/other-scripts/vis_layout.R
chreman: indeed

server/preprocessing/other-scripts/vis_layout.R
```diff
@@ -18,130 +18,189 @@ vis_layout <- function(text, metadata, max_clusters=15, maxit=500, mindim=2, max
...
+
   return(list(tdm_matrix = tdm_matrix, metadata_full_subjects = metadata_full_subjects))
 }
 
+replace_keywords_if_empty <- function(corpus, metadata) {
+
+  missing_subjects = which(lapply(metadata$subject, function(x) {nchar(x)}) <= 1)
 
+  candidates = mapply(paste, metadata$title[missing_subjects], metadata$paper_abstract[missing_subjects])
+  candidates = lapply(candidates, function(x)paste(removeWords(x, stopwords('english')), collapse=""))
```
chreman: language is now a parameter, but within function-scope, not global. may need improvement

pkraker:
Great work! I did a first test run with the default test cases, here are my observations:

- Labels for the DOAJ test case (marketing) are looking really good
- Labels for the Pubmed test case (health) are looking good too, but they do have a few problems:

1. One of the labels is "Among", which is a stopword. This should be fixed before merging.
2. A few of the labels are nonsensical: "Team team" and "Life women". We can probably never avoid labels like this with auto-generated n-grams, but maybe there is a way to reduce them (e.g. leave stopwords in for n-gram generation, but make sure that n-grams don't start or end with a stopword).

chreman:
1) I had a look at stopwords('english') from `tm`, and 'among' is not among them (pun not intended ;) ) but if we come across more cases like these (could run it through our experiment pipeline) we can add them via a custom stopword list.
2) We can leave stopwords in for n-gram generation, and then filter for n-grams containing stopwords. although this is a many-to-many comparison, it should still be acceptable for these corpora. will give it a shot.

pkraker:
ad 1. :D My bad! I thought it was using the stopword list in `resources/english.stop`, but that one is used in a PHP naming procedure.
ad 2. I think that this would be a good procedure for 2-grams. For 3-grams, it is probably sufficient to filter out those that have a stopword as the 1st or the 3rd word.

Bubblbu:
Awesome work :+1: 

Quickly skimmed the changes and realized that we're still using some very generic variable names. Most of them are still leftovers from the early days (\*cough\* wasn't me... :D), but I think it would be a good idea to force ourselves to stick to more expressive variable names, especially as an effort to improve the speed and ease of working into Headstart.

Every suggestion obviously subject to further discussion:

----

`vis_layout`

I think the general name of the file and the main routine is a bit misleading, as it is not only dealing with the ordination, but the overall map.

----

```
result <- create_tdm_matrix(metadata, text);
metadata_full_subjects = result$metadata_full_subjects
  
print("normalize matrix")
normalized_matrix <- normalize_matrix(result$tdm_matrix);
```

Minor thing, but `create_tdm_matrix` creates a list that contains more than the tdm_matrix. Maybe we can simply add a specific assignment for `tdm_matrix` (as `metadata_full_subjects`) and rename the function to `create_tdm_and_metadata` (or something similar)

----

```
distance_matrix_2 <- as.matrix(proxy::dist(tdm_matrix, method))
distance_matrix = as.dist(distance_matrix_2)
```

Maybe use tmp_dist_matrix?

chreman:
re: stopwords
Added reading additional stopwords from resources in, although the given path needs to be adapted, depending from where the script is called.
I also changed the bigram-generation in `create_cluster_labels` to add them only from titles. I also tested n-gram-generation with stopwords still in, but they weren't very satisfactory and only polluted the results. the "Team team" label comes from a cluster of two papers, where one has the title "'The team, the team, the team': that's what managing population health is all about for Matthew Davis. Interview by Stacy Sellek." oh well difficult to guard against that ;)

re: variable/function naming: agreed, for me it's also not intuitive. this is certainly something where we could come up with some new conventions. I'm open for suggestions.

pkraker:
Thanks, will take a look at this later on!

Regarding the "Team team" label: the heuristic proposed above would filter out exactly this case. I've successfully used this approach in an earlier routine. The steps are:
1. Create ngrams with stopwords included
1. Filter out all ngrams that have a stopword as first or last word
1. Filter out all ngrams where any word that is *not* the last word ends on puctuation (for example any 3-gram where the last character in the first or second word is a colon, semicolon, exclamation mark etc)

chreman:
added a filtering routine for ngrams generated from titles

pkraker:
Great work! Merging now...

Issue #106, opened 2017-02-20
vis_layout.R fails for PubMed query "russian"


pkraker:
Error message: `Error in data.frame(x.min) : row names contain missing values`

To reproduce, checkout branch `fix-vis-layout` and execute `server/preprocessing/other-scripts/test/pubmed-test.R`

pkraker:
This boils down to an item, which does not seem to have any metadata but an abstract as reported by `get_papers` in `pubmed.R`. In my result set, it's currently at index no. 26.
When searching for the abstract, however, I found a seemingly well described entry: https://www.ncbi.nlm.nih.gov/pubmed/24851295

@sckott: do you have any insights on that?

sckott:
any idea where the error is arising from in the `pubmed.R` script?  looking at it

sckott:
@pkraker  The error is in the `vis_layout.R` file - in the `create_ordination` function, and inside that in the `ecodist::nmds.min` function

sckott:
in `pubmed.R` - we assumed that all records from pubmed via `rentrez` were of type `PubmedArticle`, but some are books of type  `PubmedBookArticle`, so we can add the bits:

```r
book_fields <- c('.//BookDocument/Book/BookTitle', './/BookDocument/PMID',
                   './/BookDocument/Book/Publisher/PublisherName', './/Abstract',
                   './/BookDocument/Book/PubDate/Year')
...
flds <- switch(
   xml2::xml_name(z),
   PubmedArticle = fields,
   PubmedBookArticle = book_fields
)
...
tmp <- stats::setNames(lapply(flds, function(w) xml2::xml_text(xml2::xml_find_all(z, w))), fnames)
```

i could PR, but don't know if you want to exclude books / only have journal artcles, in which case a different unknown change could be made

pkraker:
Thanks for the quick reply. Your proposed changes look fine! Def. in favour of including books by default. Users can deselect them in the options, if needed. Please PR!

sckott:
@pkraker will do, coming soon

sckott:
do you want PR against `fix-vis-layout` or `master`?

pkraker:
`fix-vis-layout` please, this branch contains the test that reproduces the error.

sckott:
okay, will do

PR #112, opened 2017-02-23
fix to pubmed.R to handle books

sckott:
#106


pkraker:
Works very well, thanks!

PR #113, opened 2017-03-06
checking for missing methods and default values per localizations,

rbachleitner:
some fixes concerning list-only mode.

### DIFF
--- a/vis/js/default-config.js
+++ b/vis/js/default-config.js
@@ -107,7 +107,18 @@ var config = {
             title: "title",
             default_title: 'Overview of <span id="num_articles"></span> articles',
             area: "Area",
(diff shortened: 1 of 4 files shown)


pkraker:
Looks good, but there is one issue with these changes: it's not possible to remove metrics using the flag `content_based` any more. You can test this using `search_doaj.html` in the `search_repos` example.

rbachleitner:
@pkraker Can you check again? I think it was just one wrong conditional

pkraker:
Sorry, this was actually working before the requested change. Can I ask you to revert the last change?

rbachleitner:
@pkraker Should be like before again.

PR #114, opened 2017-03-07
Integration of clusternaming

Bubblbu:
Changelog:

+ Merged master with recent fixes into clusternaming
+ New labeling algorithm works for all PubMed, DOAJ & BASE.
+ Excluded articles with empty abstracts for PubMed, DOAJ & BASE


server/preprocessing/other-scripts/vis_layout.R
```diff
@@ -17,7 +17,8 @@ debug = FALSE
 vis_layout <- function(text, metadata, max_clusters=15, maxit=500, mindim=2, maxdim=2, lang="english") {
   stops <- stopwords(lang)
   if (lang=="english"){
-    additional_stops <- scan("../../resources/english.stop", what="", sep="\n")
+    # relative path is problematic as file is called from different sources
```
pkraker: This should be fixed before merging. I suggest passing the path as an additional parameter. That way, all path issues are sorted out in the main script.

Bubblbu: I also had the problem with single-word labels. Could resolve it with @chreman in a call...

I assume you recently updated your R-packages. Including the tm-package... We realized that tm 0.7-* breaks the labeling without throwing any errors/warnings. So downgrading to 0.6-* should resolve that problem. 
I updated the server requirements: 1e5f933

Agree with resolving the path issue before merging back into master!

Bubblbu:
I also had the problem with single-word labels. Could resolve it with @chreman in a call...

I assume you recently updated your R-packages. Including the tm-package... We realized that tm 0.7-* breaks the labeling without throwing any errors/warnings. So downgrading to 0.6-* should resolve that problem. 
I updated the server requirements: 1e5f933

pkraker:
Ah sorry, I didn't see that. Will try downgrading tomorrow.

pkraker:
Downgrading tm did indeed fix the problem with single keywords. Looking very good now! Once the issue with the stopwords file is fixed, we should be ready to merge :)

chreman:
opened a #115 PR for making stopwords file a parameter in vis_layout

pkraker:
I am getting now errors from line 267 in `vis_layout.R`:
```
Error in grepl(ngram, top_names) :
invalid regular expression '4-triazolo[1', reason 'Missing ']''
```

R version is 3.3.2 with all libraries but `tm` up-to-date. This error occurred in earlier tests with @chreman, but vanished after updating R. It is most likely due to the fact that ngrams contain special characters, which are interpreted as regexes. Potential fixes are escaping all special characters with yet another regex, or using a string matching function that doesn't use regexes.

pkraker:
It seems that problems with grepl can be avoided by using the stringi package, which has a non-regex string matching function. The following code worked for me in `filter_out_nested_ngrams` (it seems that there are empty ngrams, which `stri_detect_fixed` cannot deal with, hence the check):

```
for (ngram in top_ngrams) {
    if (ngram == "")
      next;
    
    ngram_in_top_names = stringi::stri_detect_fixed(top_names, ngram)
    top_names_with_ngram = sapply(top_names, function(x)(stringi::stri_detect_fixed(ngram, x)))

[...]
}
```

Bubblbu:
Fix for rel.paths & additional stops ready for re-review:

`vis_layout <- function(text, metadata, max_clusters=15, maxit=500, mindim=2, maxdim=2, lang="english",  add_stop_words=NULL, testing=FALSE)`

+ `lang` - defines language for base stopwords. *Always* loaded
+ `add_stop_words` - define name of additional stopword file in `resources/` (we can also make this one default to 'english'
+ `testing` - only set in test cases and causes the paths to be pretended with "../"

pkraker:
Works now for both test and production, thanks!

PR #115, opened 2017-03-09
made stopwordsfilepath a parameter

chreman:
path to additional stopwords file now a parameter in vis_layout

### DIFF
--- a/server/preprocessing/other-scripts/test/doaj-test.R
+++ b/server/preprocessing/other-scripts/test/doaj-test.R
@@ -32,6 +32,6 @@ input_data = get_papers(query, params)
 #time.taken <- end.time - start.time
 #time.taken
 
-output_json = vis_layout(input_data$text, input_data$metadata, max_clusters=MAX_CLUSTERS)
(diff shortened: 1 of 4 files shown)


pkraker:
This change needs to be reflected in the main scripts in production and test.

chreman:
@Bubblbu can you point me to the main scripts in production? only know about test

Bubblbu:
Sry for backpedaling, but wouldn't it be nicer (and simpler) to use a flag for the the testing case?

pkraker:
All path issues are currently handled in the main scripts and I think we should keep it that way. The approach has the added benefit that you can e.g. test different stop word lists using the evaluation framework.

Bubblbu:
@chreman for production `other-scripts/text_similarity.R` calls vis_layout()

Bubblbu:
@chreman had some time and added an implementation to PR #114 with commit 216feef

chreman:
reflected it in production and test scripts, but still as a relative path.

chreman:
closed after https://github.com/OpenKnowledgeMaps/Headstart/pull/114/commits/216feef9f2793fa235818949e48017be0d61ddcd

Issue #118, opened 2017-03-16
rbace updates


sckott:
been updating `rbace` - current version on master:

- now has fxns for the two other routes on their API, repositories and profiles. 
- added facet params and output to `bs_search()`
- output of `bs_search()` is now a list to accommodate optional facet output, so just need to change to `x$docs` to get to the same data.frame as before
- `boost` param in `bs_search` is now `boost_oa` and is a logical not a character input

pkraker:
New `rbace` version has now been integrated in Headstart.

sckott:
thanks

PR #119, opened 2017-03-16
update bs_search output

chreman:
#118 point 3: 
> output of bs_search() is now a list to accommodate optional facet output, so just need to change to x$docs to get to the same data.frame as before


pkraker:
Perfect, thanks!

PR #134, opened 2017-06-25
Backend simplification

chreman:
Split up vis-layout into different modules along functionalities:
* preprocessing, clustering, labeling
* moved APIs to separate folder
* vis_layout remains as driver of workflow


pkraker:
Tested with all available tests,. Works well, thanks for the effort!

PR #135, opened 2017-06-25
Revert "Backend simplification"

pkraker:
Reverts OpenKnowledgeMaps/Headstart#134


pkraker:
Whoops, too early. Just noted that text_similarity.R has not been updated to reflect the new structure. A search from the client is thus failing at the moment.

PR #137, opened 2017-06-25
reflected restructuring

chreman:
* api-scripts relocation now reflected in main file


pkraker:
Thanks! I just found out about the `chdir`, a magic flag for `source`. I'll take a look and see whether it solves our problems.

Issue #171, opened 2017-12-13
Area title cannot be selected and copied


pkraker:
Feedback from scavenger hunt participants. When a bubble is zoomed in, the area title cannot be copied.

pkraker:
Fixed by @jaels

PR #172, opened 2017-12-16
issue #171 - added margin-top to headstart-chart

jaels:
Hey!
Let me know if this solution is good enough.

Cheers


pkraker:
Hi, thanks for the PR! Vertical screen space is the most precious, therefore I'd prefer a solution that works without lowering the map. Also, the solution only works when the topmost bubbles are zoomed, but not bubbles closer to the bottom of the screen (try e.g. zooming into "Meta Analysis" and selecting the title then). Therefore, a solution using z-index as explained in [this Stackoverflow thread](https://stackoverflow.com/questions/33313602/how-to-place-html-elements-over-inline-svg] may be better. I did a quick test in the browser and setting a z-index in combination with a position for the svg element and the heading may work.

jaels:
Ok! good to know about the vertical delicacy. z-index was my first guess, played with it a bit, but probably not enough. Will give it another try. Thanks!

jaels:
Hey,
Sorry, accidentally pushed before pulling so there was a conflict in the css file. I resolved it. 
I think I fixed the problem with z-index and position. Have a look...
Good day!

pkraker:
Thanks, this is a better solution. However, there are two problems left:
1. The branch is still reported as having conflicts
1. The map does now overlay the list, when zoomed in. This is different to the current behaviour, see https://openknowledgemaps.org/educational-technology/

pkraker:
In addition, I realized that the whole div containing the title is now elevated. But this also means that the whole area does not react to click any more, even if it does not contain any text. This could be confusing for users, who want to zoom out or zoom into another bubble, but accidentally click on an area covered by the subdiscipline_title div. Therefore, I suggest to elevate only the actual text (which may require an additional span around the title text).

jaels:
So now the map doesn't overlay the list, and only the text itself gets the z-index. Let me know if it's fine. Notice - in canvas.js you "import dateFormat from 'dateformat'", which creates an error in the compiling, because it can't find the file in the folder. I had to comment it out.

pkraker:
Thanks, it's looking very good now! I have only noticed two minor things that need to be fixed before we're good to go:

1. Please uncomment the import for `dateformat` again, as it's a dependency that we need. You most likely need to run `npm install` again to install the dependency and then it should work.
1. It should be possible to select the word "Area:" as well.

jaels:
Hey!
Dateformat is uncommented and "area" is selectable. Let me know if there is anything else to do here. 
Cheers

pkraker:
Looking good now, thanks!

Issue #173, opened 2018-01-16
Change cursor to indicate possibility to zoom in or out of bubble


tanteuschi:
At the moment the cursor doesn't change its appearance to indicate that you can zoom in or out of one bubble. The problem is that some users don't understand how to zoom in or out of a bubble. 

Suggestion: 
**1. Zoom in** 
When map is zoomed out and the user hovers over bubble, change cursor to `cursor: zoom-in;` 
![zoom-in2](https://user-images.githubusercontent.com/19272712/34997242-3a49c1f4-fadc-11e7-89dd-9f7e5ee20ac1.png)

**2. Zoom out** 
When the map is zoomed in and the user hovers over (only) white areas outside of bubble (within map area), change cursor to `cursor: zoom-out;`
![zoom-out2](https://user-images.githubusercontent.com/19272712/34997249-3cacc414-fadc-11e7-8f6b-64da1e8d40a2.png)

PR #174, opened 2018-01-31
#173 Change cursor according to click effect (zoom-in/zoom-out) 

rbachleitner:
Issues that I noticed:

-  Zoom in cursor is already triggered at the bounding box of the bubbles (still reflects what a click would do though)
- When zooming out while hovering over a bubble, the cursor goes from zoom out to zoom in (so far so good) but then if you mouse over paper it changes to the default cursor and doesnt stay the zoom in cursor. I tried to work around that but havent found the correct thing so far.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2,6 +2,7 @@
   "name": "headstart",
   "version": "1.0.0",
   "lockfileVersion": 1,
+  "requires": true,
   "dependencies": {
     "abbrev": {
       "version": "1.1.0",
@@ -13,7 +14,11 @@
       "version": "1.3.3",
       "resolved": "https://registry.npmjs.org/accepts/-/accepts-1.3.3.tgz",
       "integrity": "sha1-w8p0NJOGSMPg2cHjKN1otiLChMo=",
-      "dev": true
+      "dev": true,
+      "requires": {
+        "mime-types": "2.1.15",
+        "negotiator": "0.6.1"
+      }
     },
     "acorn": {
       "version": "3.3.0",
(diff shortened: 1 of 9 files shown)


tanteuschi:
I have tested this in Firefox latest version. Most base cases work well - thank you!

Regarding your issues: 
no1 please clarify - I haven't noticed anything
no2 yes that should be fixed

issues I noticed (I will continue a count so we know which one we are talking about)
no3 when you are zoomed into a bubble the cursor indicates zoom-out (inside of bubble) - which it should not do. (just plain default mouse is fine)
no4 when your are zoomed in, hovering over neighboring bubble should also indicate "zoom-in" (right now it's default mouse) - -> not sure if many people use this feature (because bubble titles are missing - also sth. we should fix one day;) - but just to be consistent let's use the zoom in

This pic shows all the hover states that should occur:
hover-states

rbachleitner:
New commits with a better solution then before!

Issues No1 - No4 should be correctly addressed. Please test and give feedback!
There's still some touching up for me to do, so even if satisfied with the PR please don't merge until tomorrow evening.

While trying to make it work I discovered a new bug in the current master branch (havent looked through the issues it if its there, will do, but this is a note to myself for now):
When zooming in and then zooming out again, some of the foreignObjects + their paper_holder divs
retain the width & height of their zoomed in version. This in turn makes it impossible to hover over the bubble circle area in between papers sometimes and thus makes the mouseClick listeners there unavailable (e.g. sometimes it's impossible to zoom in by clicking there).
I worked around it, which is fine for the mouse cursor, but the clicking behaviour should be fixed.

tanteuschi:
This is looking mighty fine. If this was blossom I'd say ship it to 🎉. Great job - thank you!

pkraker:
Also looking good from my side! Only waiting on your "go" to carry out the merge.

Bubblbu:
Didn't work with Firefox for Ubuntu 58.0.2 (Ubuntu 16.04). Same applies to @chreman 

Chrome and Brave on the same machine were ok!

rbachleitner:
You can merge it. I only think I have one unnecessary cursor: zoom-in definition in the css, but I can streamline that at a later point.

rbachleitner:
Unfortunately it works on my firefox 58.0.2 so I can't chime on that now.
Will google that issue later.

pkraker:
Merging this now. @Bubblbu @chreman if the problem on Ubuntu/FF persists, please open an issue!

PR #175, opened 2018-02-01
Dragging pdf preview

jaels:
So, I used Rainer's link in stackoverflow, and now the pdf-modal is draggable without bugs. 
There is also an image-modal (images_modal.handlebars) that is supposed to be draggable now (not sure what it's used for, let me know...). Will now move on to try making it resizable


pkraker:
Thanks! The images_modal can be used when you have images of papers instead of the full pdf. You can activate this mode by setting the `preview_type` to `image` in `examples/local_files/data-config.js`.

The base case works very well now (also for the images_modal, I checked), but there are two things that I noticed:

- The modal does not close when clicking on the gray area outside the modal. This, however, is the current (and preferred) behaviour. Please re-enable this.
- You can drop the modal in the address bar of the browser, which in my case triggers a Google search with the coordinates of the modal. I imagine that this would be rather confusing for users. Can this be prevented?

jaels:
Hey Peter,
So...
- The close bug - it's solved, but now - if you just open the modal and don't move it, its all the same (click and close), but if you move it, you need to click twice outside to close it. Not sure if it's solvable - got any idea?
- Was trying to solve the address bar problem for quite some time and couldn't manage so far, I continue trying...
- About the resize issue - it's pretty complex somehow. What works perfectly in Firefox, doesn't work in Chrome at all etc. Is it possible to install jQuery UI in the project? It should have a simple tool for that, and also for dragging. Would be good at least to try. Let me know what you think

pkraker:
Thanks for all the effort you are putting into this! I am not opposed to using JQuery UI, especially if it's much simpler to go that way. The key will be to keep its size down as it is a fairly large library (but customizable AFAIR).

Regarding the address bar problem: I found [this fiddle](http://jsfiddle.net/dave_wilton/a3gyq/5/), where the problem does not seem to come up.

jaels:
How do I keep it's size down? 
I think the fiddle that you sent uses jQuery UI. It doesnt work in the code, i've tried before ("draggable is not a function"), and from my search in the web the draggable feature is always attributed to UI.

pkraker:
With jquery-ui, you can customize the components that are included (see the [Download Builder](https://jqueryui.com/download/)). I don't know if that's possible with npm though, so you might need to include it in the lib-directory.
As for the example, you are right, it uses jQuery UI as well. My bad, didn't realize that.

jaels:
Hey,
So - draggable and resizable functionality works, using a minimize version of jquery-ui. The only thing is the tiny resize logo, that should be in the right bottom corner. I cant get to see this image, which is in the images folder, or any other image from there that I try to attach, no matter what I do. The path is good, but all I get to see is blank square or a generic symbol of an image. 
Let me know what you think...

pkraker:
Thanks! Good job reducing the size of jquery-ui (only 49kb!). Also, the dragging part works indeed very well now. My last concern would be to change the cursor to `grab` (and `grabbing` when clicked) as desribed [in this post](https://stackoverflow.com/a/18294634/4061929).

Regarding resizing, I am seeing the resize cursors. You may need to set your `publicPath` in `webpack.config.js` to the HTTP path to your `dist` folder (e.g. http://localhost/headstart/dist/). There are still some problems with the resizing in my view:

* Resizing only works on the right hand side of the modal.
* Resizing on one side expands both sides of the modal. I would expect it to only expand the side I am dragging on.
* De-sizing does not work well for me. The modal does not become smaller, and it seems to be stuck in resizing mode after releasing the mouse.

Moving forward, as this PR has already become rather big, my suggestion would be to finish the work for dragging the modal in this PR, and then open a separate PR for the resizing functionality. What do you think?

jaels:
Sure, am I supposed to do that or you? (sorry, don't have much experience with that)

pkraker:
That would be your task. My suggestion would be to move all code regarding the resizing functionality to a new branch /based on this branch) and open a new PR.

jaels:
Hey, 
So, only the dragging functionality is here, and I added the cursor.
Let me know what you think...

jaels:
Hopefully the dragging is good now

jaels:
Just deleted an unwanted image file

pkraker:
Thanks, looking very good!

PR #182, opened 2018-02-16
Resizing the modal

jaels:
Hey Peter!
So...

1. "Resizing only works on the right hand side of the modal" - isnt it always like that? 
https://jqueryui.com/resizable/
Or maybe I misunderstood you.

2,3. Because the pdf modal consists of so many layers and elements - pdf-modal, modal-content, iframe-modal, pdf-iframe, modal body, viewercontainer, viewer and so on, its very hard to control and get to a perfect resizable functionality (the image modal is much simpler, and that's why it works very well).
One problem is (its a problem also in the image modal) - the element that causes all sides to resize at once is the #pdf-modal/#image-modal, but if I dont include it in the resize list, it stays the same size in the background and then when the modal gets smaller, and you think you click outside of it to close, you dont, and it doesnt close. I didn't include it now, so you can see. 

Need to decide what to do with it. Looks like its either that or that. 

Apart from that, because there are so many elements inside the pdf-modal, I believe that the resizing can't be super fast and smooth, although on my computer it works well now.

Firefox:
Does a lot of problems there. Because, for some reason, the active pseudo class for grabbing wasn't working in firefox, I hard coded it (if you were wondering). So now the grabbing cursor works in FF too.

The continuous resizing - also happens only in FF. Seems like only in the pdf-modal, and not the image-modal. Couldn't solve it so far. Let me know if you have any idea.

Waiting for your input...

Good day!


pkraker:
Closing this for now, as we will revisit this issue as part of the frontend rewrite.

Issue #131, opened 2017-06-06
Move publicPath to a separate config file


pkraker:
The publicPath property in webpack.config.js needs to be set to the current path. This should be moved to a separate config file to avoid accidental commits and git conflicts.

PR #192, opened 2018-02-23
issue #131 - move publicPath to separate config file

jaels:
People will need to change also the path to the config file from the webpack file - remove example

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ vis/stylesheets/*.css
 *.map
 dist/
 .idea/
+config.js
--- a/README.md
+++ b/README.md
@@ -7,12 +7,16 @@ Head Start is a web-based knowledge mapping software intended to give researcher
 ## Getting Started
 
 ### Client
-In order to get started quickly, Make sure to have `npm` version 3.10.10 installed (it comes with Node.js 6.12.0, you can [download installers here](https://nodejs.org/dist/latest-v6.x/)) and run the following two commands to build the Headstart Client. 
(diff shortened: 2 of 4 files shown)


tarrow:
I might have missed something but won't this simply mean we get people now accidentally submitting patches that look like?

```
-const config = require('./config.example.js');
+const config = require('./config.js');
```

pkraker:
Looks really good, but I agree with both comments by @tarrow. 

It should be:
```
const config = require('./config.js');
```

Also, please add this step to the documentation here: Readme.md#getting-started

jaels:
Hope it's good now.
Let me know...

pkraker:
Looking good, thanks! I have two small last change requests: 

1. Please change the path in `config.example.js` to `http://localhost:8080/dist/`. This way, it will work out of the box for people running the webpack server once they have copied and renamed the example file.

1. Please change README.md from
```
Open a new file named `config.js` in the main `Headstart` folder, looking like `config.example.js`, and change the publicPath to the path of the dist directory.

Dev: specify a full path including protocol (it's usually the same path specified in the example).
Production: specify full path excluding protocol e.g. `//openknowledgemaps.org/headstart/dist`

Now you can run a local dev server: 
 	npm start
```

to 

```
Duplicate the file `config.example.js` in the root folder and rename it to `config.js`. 

Now you can run a local dev server: 
 	npm start

To run Headstart on a different server (e.g. Apache), you need to set the publicPath in `config.js` to the URL of the `dist` directory:
* Dev: specify the full path including protocol, e.g. `http://localhost/headstart/dist`
* Production: specify the full path excluding protocol, e.g. `//example.org/headstart/dist`
```

jaels:
Done.

pkraker:
Awesome! Merging now :+1:

PR #193, opened 2018-03-02
German preprocessing

chreman:
* added stopwords to resources
* added/adapted language flags for stopwords and stemming
* language is now a command line argument in text_similarity, which defaults to english if not given


chreman:
works in local staging, but creates some ugly artifacts: missing Umlaute, lowercasing of nouns

pkraker:
Thanks! A public test would be very beneficial for this PR. My suggestion would be to add an option to BASE to search for documents in German only. For this to work, you have to add `dclang:ger` to the query string.

chreman:
* added a language flag in the text_similarity call of get_papers
* reflected that flag in all get_papers methods
* it is only used in base so far, where `dclang:ger` is added to the query string if called

chreman:
Thanks for the input, indeed it makes more sense for language to be passed in the params object. I removed it from the get_papers where language is currently not an expected functionality.
Encoding errors for me did not appear in Rstudio when prototyping/testing, rather on the final map of the full pipeline. for this I did a local staging deployment, but tweaked the base get_papers and vis_layout to default to 'german'

server/preprocessing/other-scripts/text_similarity.R
```diff
@@ -53,14 +52,25 @@ if(!is.null(params_file) && !is.na(params_file)) {
...
+  } else {
+    language <- 'english'
+  }
+
+ADDITIONAL_STOP_WORDS = language
 print("reading stuff")
 print(params)
 
-input_data = get_papers(query, params, limit=limit)
+input_data = get_papers(query, params, limit = limit, language = language)
```
pkraker: This line triggers an error:
```
Error in get_papers(query, params, limit = limit, language = language) :
  unused argument (language = language)
Execution halted
```

server/preprocessing/other-scripts/text_similarity.R
```diff
@@ -53,14 +52,25 @@ if(!is.null(params_file) && !is.na(params_file)) {
   params <- NULL
 }
 
+if ('language' %in% params){
+    language = params$language
```
pkraker: Here you have to take into account that the frontend may send "all" for all languages. In this case, my suggestion would be to use also set the language to `'english'`.

server/preprocessing/other-scripts/base.R
```diff
@@ -23,34 +23,43 @@ library(rbace)
...
 
   year_from = params$from
-
   year_to = params$to
-
   date_string = paste0("dcdate:[", params$from, " TO ", params$to , "]")
+  if ('language' %in% params){
+      language = params$language
+    } else {
+      language <- 'english'
```
pkraker: The default should be no filter, not english.

chreman:
* Params and language defaults cleaned up
* fixed some bugs
* reverted accidental commits of openaire-code

both tests for default and base_ger now run through

chreman:
I'm sorry for the long delay. The last commit should address your required changes.
The issue with the unintended lower casing of keyword remains, but I'd suggest to treat it as a separate issue.

pkraker:
Excellent, thanks! Changes work as described.

PR #199, opened 2018-03-13
Elbow k fix

### LINKED ISSUE #169: Elbow heuristic fails for PubMed query with useful results
The currently checked-in test for PubMed fails, even though the search yields useful results (query: "russian", sorting: "most-recent", from:"1809-01-01", to:"2017-12-04",article_types:all)

The elbow heuristic does not complete with the following message:

````
 Error in elbow.batch(css_cluster) : 
  A good `k' is not available with provided inc.thres and ev.thres; please make adjustment, e.g. decrease `ev.thres', increase `inc.thres' or increase `k'. 
````

chreman:
added error handling and a clustering-fallback with drastically changed parameters
should fix #169 at least for the test case


pkraker:
Worked well in all my test cases - great work! Merging this now.

PR #201, opened 2018-03-15
VIPER Examples and Backend WIP

tarrow:


pkraker:
Thanks, it's looking great! I found only one minor issue with the backend connection:

* The query you send is the actual user query (e.g. "sugar"). But it should rather be the selected project's acronym (e.g. "SURUCO"), so that the map then says "Overview of SURUCO..."

I know that you are still working on the example, but here's my input: I think that it should give a bit more feedback to the user (similar to the other examples), otherwise you are lost without tracking the network requests. This includes:

* The submit button should be inaccessible once it's been clicked and the search is running (see other examples).
* When the ajax call for the search is running, some sort of indicator should be shown (e.g. the loading bar from the other examples).
* Once you click on "Make map", a new tab should be opened with some sort of indicator. If it succeeds, the map should be shown. If it fails, an error should be shown.

pkraker:
Thanks, it's easier to work with now, but there are still a few issues:
* After I have searched and the results are returned, the search bar stays grayed out and cannot be used again.
* When I click on "Make map", the list disappears - it should rather be disable for the time of map creation.
* When the map creation fails, it correctly displays `Sorry! Something went wrong. Most likely, we did not get enough results for your search. Please try again with a different query.`, but the loading indicator and the phrase `Building OpenKnowledgeMap` remain. The list does not return. When I click on search again, I get the following warning in an alert box: `DataTables warning: table id=tbl-project-search-results - Cannot reinitialise DataTable. For more information about this error, please see http://datatables.net/tn/3`

tarrow:
Aborting this PR.

The backend work from here has already made it in. More backend and frontend to come in separate PRs.

PR #205, opened 2018-03-15
Openaire - now XML response based

chreman:
test runs

### DIFF
--- a/server/preprocessing/other-scripts/openaire.R
+++ b/server/preprocessing/other-scripts/openaire.R
@@ -59,21 +59,21 @@ get_papers <- function(query, params, limit=NULL) {
   datasets_md <- datasets$response$results$result$metadata$`oaf:entity`$`oaf:result`
 
   tryCatch({
-    pubs_metadata <- build_pubs_metadata(pubs_md)
+    pubs_metadata <- parse_publications(pubs_md)
     pubs_metadata$id <- pubs$response$results$result$header$`dri:objIdentifier`$`$`
   }, error = function(err){
     print(err)
-    pubs_metadata <- data.frame(matrix(nrow=1))
   }, finally = {
+    pubs_metadata <- data.frame(matrix(nrow=1))
   })
 
   tryCatch({
-    datasets_metadata <- build_datasets_metadata(datasets_md)
+    datasets_metadata <- parse_datasets(datasets_md)
     datasets_metadata$id <- datasets$response$results$result$header$`dri:objIdentifier`$`$`
   }, error = function(err) {
     print(err)
-    datasets_metadata <- data.frame(matrix(nrow=1))
   }, finally = {
+    datasets_metadata <- data.frame(matrix(nrow=1))
   })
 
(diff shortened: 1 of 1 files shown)


pkraker:
Awesome, thanks! It works well for certain calls (see e.g. project ID 246686, 611742), but fails for others. Here are the two errors that I've seen, along with relevant project IDs. I suspect that most of them do not have enough related publications/datasets.

Project IDs: 230636
Error:
```
> input_data = get_papers(query, params)
Warning in xpath_search(x$node, x$doc, xpath = xpath, nsMap = ns, num_results = Inf) :
  Undefined namespace prefix [1219]
xmlXPathEval: evaluation failed
```

Project IDs: 911723, 8488
Error:
```
> input_data = get_papers(query, params)
Warning in xpath_search(x$node, x$doc, xpath = xpath, nsMap = ns, num_results = Inf) :
  Undefined namespace prefix [1219]
xmlXPathEval: evaluation failed
Warning in xpath_search(x$node, x$doc, xpath = xpath, nsMap = ns, num_results = Inf) :
  Undefined namespace prefix [1219]
xmlXPathEval: evaluation failed
Error in matrix(nrow = nrow(all_artifacts)) : data is too long
```

pkraker:
I also noticed that project ID 330731 reports 67 publications with `roa_pubs`, but only 22 are reported in the visualization. Here is the relevant output from that run:

```
> input_data = get_papers(query, params)
Warning in xpath_search(x$node, x$doc, xpath = xpath, nsMap = ns, num_results = Inf) :
  Undefined namespace prefix [1219]
xmlXPathEval: evaluation failed
> 
> output_json = vis_layout(input_data$text, input_data$metadata, max_clusters=MAX_CLUSTERS,
+                          add_stop_words=ADDITIONAL_STOP_WORDS, testing=TRUE, list_size=-1)
Read 574 items
[1] "calc matrix"
[1] "normalize matrix"
[1] "create clusters"
Warning in weightSMART(x, spec = "ntn") : empty document(s): 10
Minimum stress for given dimensionality:  0.2941957 
r^2 for minimum stress configuration:  0.636748 
```

chreman:
improved error handling, so it now returns NULL and not a total crash. need to look a bit more into the specific projects to make sure it is always related to the low/zero number of results and not something else.

pkraker:
Errors are looking much better now, thanks! The problem with project ID 330731, however, is still there (see above).

chreman:
I can reproduce the error on my part, but currently I'd have to relay it to openaire-API.
`roa_pubs(fp7 = "330731")` indeed returns 67 publications, on the other hand `http://api.openaire.eu/search/publications?projectID=330731&funder=EC&format=xml` for the XML-format the API shows me `<total>22</total>`

pkraker:
Thanks! So this is a bug then that we should report back to OpenAIRE. From my point of view, this can be merge. But please revert the last commit - we always report a more precise date, if we can.

pkraker:
Excellent, merging this now!

PR #206, opened 2018-03-19
Viper dataset item type

tarrow:
New item type for datasets which is a rounded rectangle.

Also includes slight changes to the item/paper list so that links are shown and OA ones are in red/orange. Also never shows PDF links on dataset items


pkraker:
Thanks, it's looking really good! Displays well for me after adding a column resulttype to the csv. Zooming works like charm, same goes for resizing (both onmouseover and when resizing the window). 

The only thing that is missing from my point of view is the little font-awesome icon before the title in the list (signifying either a publication or a datatset). Here are the two codes:

```
<i class="fas fa-database"></i>
<i class="fas fa-file-alt"></i>
```

tarrow:
Final commit adds the icons to the list.

pkraker:
Looks great, thanks! Merging this now.

PR #212, opened 2018-03-26
Openaire-search

chreman:
* now allows for combined search of funder and project_id (grant ID)
* incorporates latest updates from rOpenaire


chreman:
* merged upstream/openaire
* fixed the above error which occurs in the case that no altmetric can be found for any paper in a dataset

pkraker:
Tested it with several random projects and this works very well now, thanks!

PR #213, opened 2018-03-26
Viper metric - fix for crashing visualizations

chreman:
* the error was introduced at the stage of merging altmetrics with the existing output
* some DOIs we get from openaire look like this `"10.1108/PROG-08-2013-0045` and when we get altmetrics, the doi is returned as `"10.1108/prog-08-2013-0045"`. when we perform a merge operation on the results, some DOIs don't match and empty rows (publications) are created filled with NULL
* the fix is in line 121 of openaire.R `df$doi <- unlist(lapply(df$doi, tolower))`
* the PR also includes a merge of #212

### DIFF
--- a/doc/server_config.md
+++ b/doc/server_config.md
@@ -33,7 +33,7 @@ Make sure you have the following packages installed:
   * rbace (Currently, a Github repository only: http://github.com/ropenscilabs/rbace. Install with devtools.)
   * ropenaire (For VIPER. Currently, a Github repository only: https://github.com/sckott/ropenaire. Install with devtools.)
   * readr (for ropenair/VIPER)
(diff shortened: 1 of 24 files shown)


chreman:
Given the current functionality I wouldn't call it faulty DOIs, rather a case of different naming conventions - the openaire-API allowing uppercasing in DOIs, while altmetric returns lower cased DOIs.

I currently don't expect this to be an issue, but you're right that we should have this covered for the future. Will add a change.

chreman:
* altmetrics search results will now be merged with the dataset via a left (outer) join, which ensures that only rows will be merged that have a corresponding DOI in the output dataset

pkraker:
Thanks - tested and ready for merge!

PR #215, opened 2018-03-28
Viper Frontend Metric Scaling

tarrow:
Scaling by some metrics

Configurable in the headstart config file.

Includes dummy explanations

Depends on some frontend and back end commits that haven't been merged yet.

### DIFF
--- a/examples/search_repos/data-config_openaire.js
+++ b/examples/search_repos/data-config_openaire.js
@@ -16,7 +16,8 @@ var data_config = {
     area_force_alpha: 0.015,
     show_list: true,
     content_based: false,
-	
+	//debug: true,
 	show_context: true,
-	create_title_from_context: true
+    create_title_from_context: true,
+    scale_types: ['citation_count', 'cited_by_tweeters_count', 'readers.mendeley']
 };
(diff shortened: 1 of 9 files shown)


pkraker:
Thanks, this looks great! Worked out of the box for me when testing it with the OpenAIRE project. 

I have only a few minor comments, which all have to do with the config:

* Please move `scale_explanation` in the config to the localization (currently located in `default-config.js`). This would be a good opportunity to create an own localization for OpenAIRE (currently it's `eng_pubmed`).
* In the default mode, `content_based` should be set to true and it shouldn't be possible to sort by citations (to achieve this, simply remove it from `sort_options` in the config).
* The labels in the list, the sort options, and the papers should change depending on which metric is selected. My suggestion would be to simply use `citations`, `tweets` and `readers` for now. This can be achieved by adapting `base_unit`` and `sort_options` in the config. As for the sort_options, you may need to rename the attributes in `io.prepareData` to enable the sort functionality (e.g. `citation_count` to `citations`).

pkraker:
Contrary to what I said above in the last bullet point, you don't need to rename the attributes to make sure the sorting works - you simply need to change `readers` in the localization to the desired label.

pkraker:
One final comment: it should be possible to set different labels for the options in the dropdown in the localization. So for example set `citations` for `citation_count` and so on.

tarrow:
Thanks for the comments. I'll address them with a new patch in the next few days.

PR #217, opened 2018-03-28
Viper frontend improved examples

tarrow:
Improved frontend experience.

Still needs proper UX input but should help us in testing.

Addresses issues made in: #201 

Includes backend commits that are not yet merged. e8cefb5f939d61d5522b774fc7b29c6e1406f83a

### DIFF
--- a/examples/search_repos/search.js
+++ b/examples/search_repos/search.js
@@ -67,34 +67,55 @@ $("#searchform").validate({
     }
 });
 
-var doSubmit = function (data) {
+var doSubmit = function (data, newWindow, callback) {
   data += "&today=" + new Date().toLocaleDateString("en-US");
+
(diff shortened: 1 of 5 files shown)


pkraker:
Thanks, but I couldn't test it as the search fails for me with the following JS error in the console:
```
TypeError: deepGet(...).map is not a function
```

tarrow:
Sorry about that. Turns out there is a special case in the response json when a project only has one related organisation. I've tack on a commit to fix this edge case. I've now tried quite a few searches so hopefully I'm not missing any.

pkraker:
Thanks! This looks good now apart from a few minor problems, for which I will open an issue. Merging this now in order to resolve some of the dependencies between PRs.

PR #219, opened 2018-04-03
Viper more info modal

tarrow:
Change more info modal to suit VIPER projects and include their details.

Possible ToDo:  Sort out `info` vs `intro` both are used to mean basically the same thing. Might be clearer if we stuck to one. I generally stuck to `info` unless it seemed to conflict with what was there.


pkraker:
Thanks! This solution has all the required fields, but it is missing the link to OpenAIRE. Minor remark: for `Open Access Pilot` and `Special Clause` it should state `yes` or `no` instead of `true`and empty.

PR #221, opened 2018-04-03
Modals for OpenAire Edit and Embed

tarrow:
Two new modals for embedding and 'adding or changing openaire project files'

ToDo: Embed modal really needs configuring with a production base path. Could be either coded into the config or taken from from window.location. I guess we might want to use a special endpoint to quantify usage (and see referers?)


pkraker:
Thanks! This is working as expected - merging now.

PR #226, opened 2018-04-27
Rolling k

chreman:
* for small n_papers (n <=30) uses a different, simple approach to setting cluster number (not elbow): `round(squareroot(n)) + 1`
* this commit also includes some subject cleaning which should help improve cluster labels


pkraker:
Nice work! This has improved the appearance of maps with fewer than 30 items considerably in my tests. Merging this now into the preview version, but ultimate approval depends on a separate review by @tanteuschi.

PR #228, opened 2018-04-30
Viper automization

chreman:
mainly for component 1:
* new query class in `server/classes/headstart/persistence`: `ViperUpdater.php`
  * returns maps created by VIPER from openaire-API, filters according to vis_changed flag
  * TODO for later: separate queries for ObjID and funder-projectID tuples
* new service in `server/services`: `updateViper.php`, a CLI-runnable script that:
  * takes options as specified
  * decides on an action (or fails)
  * retrieves maps for update according to options
  * runs search and creates new revisions - NOTE FOR REVIEW: please check validity of search params in L125-127 to make sure new revisions are indeed created

if database is set to the test database, the component can be executed from commandline by
`php server/services/updateViper.php --vis_changed false`

the other files are work-in-progress commits, and also require `php7.0-xml` (in linux)


chreman:
I added a few commits that should address your change requests for (1.), and also include an updated test database: column vis_changed added; two openaire projects present, one with vis_changed=0, one with vis_changed=1.
Running `php server/services/updateViper.php --vis_changed false` updates both maps, `php server/services/updateViper.php --vis_changed true` updates only the one where the flag has been set.

pkraker:
Great, setting the option `vis_changed` to `true` did work now. However, I would assume that the flag be reset after the revision has been created. 

The problem when creating the snapshot persists. I would assume that the only way to fix this issue would be to add a new setting to the config, as the $SERVER variable is not available when called from the command line.

chreman:
currently adding the last functionalities, this is just a notification that the incoming commits are not ready for review

chreman:
Component one has new functionalities ready for testing:
* the base scenario of update via DB-query `php server/services/updateViper.php --vis_changed true` should now work without failing at the snapshot-stage
* update by funder-project tuples are now possible, which can be provided via CSV-file, e.g. `php server/services/updateViper.php --vis_changed false --test --funderproject=funderproject.csv`
* updates by object_ids are possible in theory, also provided via CSV-file, e.g. `php server/services/updateViper.php --vis_changed true --test --object_ids=object_ids.csv` unfortunately the format of the IDs leads to an irritating SQL query problem at the moment
* the flag-reset after 24h is now also implemented

The implementation as a pure script is not very elegant at the moment, as can be seen in the passing through of for example the `$persistence` variable, we should think of creating a class for the two components - especially since they also share a lot of functionality.

webpack.config.js
```diff
@@ -14,7 +13,7 @@ const common = {
     output: {
         path: path.resolve(__dirname, "dist"),
 		//dev: specify a full path including protocol, production: specify full path excluding protocol
-        publicPath: config.publicPath,
+        publicPath: ".",
```
pkraker: Please revert this change. You can avoid errors by duplicating the file `config.example.js` and renaming it to `config.js`. Don't forget to set your public path in `config.js` to `.` then.

pkraker:
Thanks for these changes. Here are my findings from testing:
* `php server/services/updateViper.php --vis_changed true` and `false` now completes without an error. I would suggest, however, to add an additional config option that lets you set the protocol explicitely and to revert to "http:" only when this config option is not set.
* `php server/services/updateViper.php --vis_changed false --test --funderproject=funderproject.csv` does also work well. I suggest moving the example file `funderproject.csv` to `examples/viper` however.
* `php server/services/updateViper.php --vis_changed true --test --object_ids=object_ids.csv` failed for me with the following error:
```
Fatal error: Uncaught exception 'PDOException' with message 'SQLSTATE[HY000]: General error: 1 unrecognized token: ":"' in C:\Users\pkraker\xampp\htdocs\headstart_master\server\classes\headstart\persistence\ViperUpdater.php on line 60

PDOException: SQLSTATE[HY000]: General error: 1 unrecognized token: ":" in C:\Users\pkraker\xampp\htdocs\headstart_master\server\classes\headstart\persistence\ViperUpdater.php on line 60

Call Stack:
    0.0010     153608   1. {main}() C:\Users\pkraker\xampp\htdocs\headstart_master\server\services\updateViper.php:0
    0.0090     362464   2. getCandidatesByObjectIDs() C:\Users\pkraker\xampp\htdocs\headstart_master\server\services\updateViper.php:56
    0.0090     362768   3. headstart\persistence\ViperUpdater->getUpdateMapsByID() C:\Users\pkraker\xampp\htdocs\headstart_master\server\services\updateViper.php:113
    0.0090     363488   4. PDO->query() C:\Users\pkraker\xampp\htdocs\headstart_master\server\classes\headstart\persistence\ViperUpdater.php:60
```
* I did mistakenly ask for the `vis_changed` flag to be reset after running `php server/services/updateViper.php --vis_changed true`. Please revert this change - sorry about that!

* After running `php server/services/updateViper.php --vis_changed true`, I manually set the flag for `9b34f83d3940dc249c4ae9420c8fb2e2` to 1. Running `php server/services/updateViper.php --vis_changed true` after that did not create a new revision of `9b34f83d3940dc249c4ae9420c8fb2e2` however.

chreman:
* reverted erroneous change in webpack.config.js
* moved examples CSV inputs to example folder
* removed defunct functions from ViperUpdater persistence class
* added snapshot_local_protocol option to configs and adapted checks in Snapshot class
* updates via object_id now working

* `php server/services/updateViper.php --vis_changed true`
* `php server/services/updateViper.php --vis_changed false --test --funderproject=examples/funderproject.csv`
* `php server/services/updateViper.php --vis_changed true --test --object_ids=examples/object_ids.csv`

pkraker:
Thanks - I tested the first component and it works very well. All of the test cases did run through (all tests performed with the checked in database `headstart_test.sqlite`). 

I only found one issue:

After running `php server/services/updateViper.php --vis_changed true`, the revision for vis_id:06ec94c8566c146b5b2bff761660dd79 is created, but the `vis_changed` flag is set to 0.

One further remark: `changed_timestamp` always remains at `2013-02-01 12:36:01` for both visualizations. I don't know if that's planned.

pkraker:
The functionality has been clarified, thanks! `changed_timestamp`  controls resetting of `vis_changed`. It is expected that the timestamp is set by the user. Merging this now!

PR #229, opened 2018-04-30
fixed rolling k behavior

chreman:
fixed failure in cases where no num_clusters could be detected conclusively


pkraker:
Awesome! I have tested this with a variety of projects and have not found a single one that fails.

PR #230, opened 2018-05-01
Make scale toolbar visible above the fold

tarrow:
Shrink the chart size by the height of the scale tool bar to ensure
it is visible to users on first glance.

Done by shrinking canvas.available_height consistent with the way
it is shrunk for subtitle_height.

Also refactor to remove typo in 'subtitle_heigth'.


pkraker:
Thanks, the changes are already looking great. There is only one concern that I have: there is a gap between the list and the scale div, which prevents the scale div from being shown fully (see image). Could you please remove that gap?

scale_gap

pkraker:
Nb. I merged the current version into the openaire branch for the preview.

PR #231, opened 2018-05-01
Make link and edit modal buttons clickable in zoomed mode

tarrow:
Set the modals div style to have a relative position and
z-index: 3 so it floats above the map and is still clickable when zoomed
in.

This is consistent with the fix for subdiscipline_title introduced by
ba441853


pkraker:
Works like a charm, thanks!

PR #232, opened 2018-05-01
Add2Any Share button

tarrow:
Include add2any share button using their script in a new handlebars
template for the purpose.

Append it to the modals div if config option is enabled.


pkraker:
This works very well, thanks. There is only one minor detail that needs revising: the share button is less wide than the others (see image).
addtoany_pixels

tarrow:
After a little digging it turns out that that icon was just naturally a bit smaller. They're all set to fixed width now.

pkraker:
Thanks, that solves the issue!

PR #233, opened 2018-05-01
Hide unused context elements

tarrow:
To stop some odd margins appearing that we don't want hide them
with .hide() if the config variables they are set from in canvas.js
are false


tarrow:
Its unlikely you looked at this already but I did some cheeky force pushing to fix some formatting ca 1min after making the PR.

I won't make any more after now, I'll just make a new commit.

pkraker:
Looking good, thanks!

PR #234, opened 2018-05-01
Hide dataset icons on papers on map when not a dataset

tarrow:
Short fix to now hide correct element after new id implemented for rounded border for dataset implemented.

Also: Refactor to remove commented code from template to improve clarity


pkraker:
Great, thanks!

PR #236, opened 2018-05-03
 Sort with dropdown not buttons

tarrow:
Add new config variable 'sort_menu_dropdown' with default of false
set to true in openaire config.

Add conditional section to list_explorer.handlebars. If
sort_menu_dropdown is true show bootstrap dropdown in place
of radio buttons.

in list .js rename addSortOption to addSortOptionButton for existing
functionality. Introduce new addSortOptionDropdownEntry.

Add "citations" as a sort type for viper and localisation too.

Small refactoring of template + localisation to aid clarity.

### DIFF
--- a/examples/viper/data-config_openaire.js
+++ b/examples/viper/data-config_openaire.js
@@ -14,7 +14,7 @@ var data_config = {
     show_timeline: false,
     show_dropdown: false,
     preview_type: "pdf",
-    sort_options: ["title", "authors", "year"],
+    sort_options: ["authors", "title", "citations", "year"],
     sort_menu_dropdown: true,
     is_force_areas: true,
     language: "eng_openaire",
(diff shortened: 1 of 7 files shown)


tarrow:
This took a little longer than expected. I thought I'd at least ship a work in progress so there is something to look at.

Currently not totally finished because:
- Bug meaning no initial sort label when dropdown enabled
- Dropdown is currently missing radio buttons as in sketch

pkraker:
Thanks, it's looking good so far! I assume that the left grey border of the list is only temporary, isn't it?

tarrow:
Two bullets I mentioned are fixed.

The grey bar is just from the padding, I assume that when Maxi tweaks it it will all go. I looks like before there was some specific padding for different sized screens etc..

pkraker:
Thanks, this works fine now. The only remainung issue in that the inital sort option is not selected, but that can easily be fixed in integration. Merging this now.

PR #237, opened 2018-05-06
Filter by property

tarrow:
Dropdown to filter by some property of papers

This will be superseded by a PR which also includes code to show the number of papers in the list in the header.

Just here so @pkraker can have a look at how it's going :)

### DIFF
--- a/examples/viper/data-config_openaire.js
+++ b/examples/viper/data-config_openaire.js
@@ -15,6 +15,7 @@ var data_config = {
     show_dropdown: false,
     preview_type: "pdf",
     sort_options: ["title", "authors", "year"],
+    filter_menu_dropdown: true,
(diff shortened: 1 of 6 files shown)


pkraker:
Thanks, it's looking great so far! I have tested it extensively and I have found only one sequence that causes an issue:

1. Click on a bubble not containing a dataset
1. Click on "show: Open Access"
1. Click on "show: papers"
1. Click on "show: datasets"
1. Zoom out

Now, the list and map are empty. Same goes for zooming into a different bubble.

pkraker:
Update: you can actually leave out step 2 to reproduce the issue.

tarrow:
Closing because new PR for both this and header list item count

PR #238, opened 2018-05-06
Refactoring of Filtering, Implement Filter dropdown and header item count

tarrow:
Everything in #237 but also fix for the bug @pkraker found.

Also includes header item count. This is done by working through the list, filtering for those items that "should" be in the current view and then counting those that are not marked as filtered_out: true

The implementation is rather ugly because determining what "should" be in the current view is determined after the transition.

Since I was pushed for time and couldn't see a better way to figure out either what the transition was or when it would finish it simply has a "magic" number wait.

You can therefore mess with it by transitioning very quickly.

It would be nicest to have another field in the data for example, "is_visible" which we toggle as zooming in and out.

### DIFF
--- a/examples/viper/data-config_openaire.js
+++ b/examples/viper/data-config_openaire.js
@@ -15,6 +15,7 @@ var data_config = {
     show_dropdown: false,
     preview_type: "pdf",
     sort_options: ["title", "authors", "year"],
+    filter_menu_dropdown: true,
     is_force_areas: true,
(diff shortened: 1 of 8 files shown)


pkraker:
Nice work! This does solve the above mentioned issue with filtering and implements a working version of the list count. I will take a look at the timeout solution as there are a number of flags that we can leverage to avoid having a timeout. Merging this now, after resolving the conflicts.

PR #239, opened 2018-05-06
Show tweets, readers and citations in list

tarrow:
Enabled by a config option and localisation for strings.

A span stored in a new handlebars template. Should be simple to style
or add new classes to add more detailled style if required.


pkraker:
Thanks - this works fine in default mode, but there are some issues when the metrics are switched. This will be fixed in integration.

PR #240, opened 2018-05-07
Add DOI or Link to List entry

tarrow:
add config option to enable

localise 'link'

handlebars template for each outlink span


pkraker:
Thanks, this works very well!

PR #241, opened 2018-05-07
fix scale toolbar gap

tarrow:
As you can see I just poked a single magic number.

Looks like it was originally introduced more than a year ago.

I imagine this is some function of the css and thus might need another poke after Maxi has done up the padding.


pkraker:
Thanks, this does solve the issue!

PR #247, opened 2018-05-09
Fix 243 by calling count from mediator in correct places

tarrow:
Had to add a zoomin_complete method.

Also had to set current_enlarged_paper when clicked in the bubble
not just the list.

Also unset it when clicking in current bubble

Also fire on "zoom" even if no transition has to happen.


pkraker:
Excellent work, thanks! The counter behaves now as expected.

PR #249, opened 2018-05-11
robustified elbow_k method

### LINKED ISSUE #246: VIPER: elbow fails for large projects
For large projects with more than 500 outputs (such as EC/226716 or EC/604102), the elbow fails with the following error message:

```
Error in cutree(cluster, k = num_clusters) : 
  either 'k' or 'h' must be specified
```

chreman:
fixes #246


pkraker:
Excellent, thanks! Closes #246.

PR #251, opened 2018-05-14
Make credit togglable from config variable

tarrow:
Add new config variable viper_credit.

Pass variable to scale_toolbar

Add handlebars 'if clause' to scale_toolbar.handlebars


pkraker:
Works great, thanks!

PR #252, opened 2018-05-14
Param in headstart.php for embed mode

tarrow:
if &embed=true is passed then hide the footer and set
viper_credit=true in the config.

This can then be used to display the credit image


pkraker:
Please include a check for `isset($_GET['embed'])` to avoid issues when the parameter is not set.

PR #259, opened 2018-07-24
CORE get_papers method added

sckott:
I haven't quite got the test to work yet. There's some errors internally, but they are quite deep and i'm not familiar with the vis_layout function and it's internal functions to quickly debug it. Can have a look soon, need to return to some other work now though


chreman:
Thank you for the addition! I will try to replicate the errors and hopefully can add some commits in the next days.

chreman:
@sckott I've managed to identify the error and committed a fix, the test now runs.
But it also shows that without further cleaning the maps will be very noisy. I suggest we close this PR for now and continue working in this branch until we've added additional preprocessing/cleaning measures to the pipeline.

sckott:
> Are you familiar enough with the CORE-API 

no

> I suggest we close this PR 

sounds good!

pkraker:
Awesome! Thanks so much @sckott for the integration - and @chreman for getting the test to run! 

I agree with closing this PR, but will hand this over to the frontend team so that we can better understand where and how we need to adapt the preprocessing so that it works for OKMaps.

Issue #265, opened 2018-08-22
CORE: data frame returned by get_papers is missing important fields


pkraker:
The current implementation of the get_papers method for CORE is missing the following fields from the [original specification](https://github.com/OpenKnowledgeMaps/Headstart/commit/9933bdba0f291f147163e7bbeb15acf2e64e6ecb#diff-6981f2e8882de90f3ee88264f39afe36):

* "published_in": name of the journal or venue
* "url": URL to the landing page
* "oa_state": open access status of the item; has the following possible states: 0 for no, 1 for yes, 2 for unknown
* "link": link to the PDF; if this is not available, a list of candidate URLs that may contain a link to the PDF

Issue #263, opened 2018-08-22
CORE: Implement parameters from original specification


pkraker:
The CORE connector implements different parameters than [the original specification](https://github.com/OpenKnowledgeMaps/Headstart/commit/9933bdba0f291f147163e7bbeb15acf2e64e6ecb#diff-6981f2e8882de90f3ee88264f39afe36). They were:

* from: publication date lower bound in the form YYYY-MM-DD
* to: publication date upper bound in the form YYYY-MM-DD
* article_types: in the form of an array of identifiers of article types
* langs: in the form of an array of language identifiers (["all"] for all languages)
* sorting: can be one of "most-relevant" and "most-recent"

From my point of view, there is no harm in keeping the additional parameters (the key param is even mandatory in the current implementation), but these five parameters should also be implemented to enable our usual query refinement options.

chreman:
I'm currently [working on this](https://github.com/chreman/Headstart/commit/b6f797222ab0b8885bf15a50cbbe3caf3e15cfc9), but this requires a few additions to rcoreoa. I will open a PR with [propositions](https://github.com/chreman/rcoreoa/commit/7fb4aa9395d89d261e103768e7571d847cefb61f) for additional query params soon.

chreman:
It also appears that some functionalities will have to diverge from our current ones - CORE only provides article publication date with a yearly resolution. The parameter 'repositoryDocument.metadataUpdated' may be misleading because it refers to the date when articles were added to CORE, not when they were published. Furthermore, 'article_types' is not a usable field, and sorting is only available via a relevance score. I will contact CORE for clarification.

pkraker:
Thanks! Just a quick comment regarding publication date: I have seen the field `dateUpdated`, which seems to carry the publication date in the best resolution that they have. This would be similar to how publication date is reported in BASE and should do for our use case.

chreman:
Are you referring to datePublished as returned by rcoreoa? I think this is the value we get in the returned metadata, but not what CORE allows us to search in their API.

chreman:
@sckott I attempted an implementation of a language search param in rcoreoa, please see [PR](https://github.com/ropensci/rcoreoa/pull/16). Unfortunately this does not have any effect as some of my tests with English/en, French/fr, or German/de showed - but perhaps I'm missing something in the rcoreoa - thanks for having a look!

sckott:
thanks

sckott:
@chreman did feedback

[...]

PR #268, opened 2018-08-22
Core fields

chreman:
* please note that with this commit the CORE API key is expected to be set as an Renviron variable as detailed [here](https://github.com/ropensci/rcoreoa)
* switched search behaviour to core_advanced search to enable queries as requested in #263 (search by year) - fine grained date search and other parameters are still pending
* added url, oa_state and link as requested in #265 
  * oa_state is not given in CORE metadata and defaults to 2 (unknown)
  * url and link currently follow heuristics

`published_in` may be tricky, there is only `publisher` in the metadata which looks like
```
[51] "Institute for Apprenticeships"                                      "Institute for Apprenticeships"                                     
[53] "Institute for Apprenticeships"                                      "International Council for Open and Distance Education (ICDE)" 
```
or `journals` which looks like
```
[[90]]
  title               identifiers
1    NA 2013-9144, issn:2013-9144

[[91]]
  title               identifiers
1    NA 2297-198x, issn:2297-198X
```
I'm not sure these are the journals we are looking for.


pkraker:
Thanks! Have you checked that the Renviron approach does work with the system call from PHP? For me, it works in RStudio but not when I call it from the frontend.

As we do need `oa_state` in the frontend to determine when to display the PDF link, my suggestion would be to use the PDF link as a workaround (i.e., if it exists, oa_state is set to 1, otherwise to 2).

For `pulished_in` the journals field looks like the best candidate. Can you check, if name is always N/A, and if it is, flag this to the CORE people?

chreman:
Yes, I have checked the Renviron approach. Depending on which user calls R, you may have to set it for either user, root, or Renviron.site.

I added a heuristic for the oa_state as described, and a heuristic for the published_in field, which defaults to `""` if no information is available, takes the journal title if not NA, or the identifier if present. In the test case it looks like
```
[81] ""                          ""                          ""                          ""                          "2307-3187, issn:2307-3187"
[86] "2304-070x, issn:2304-070X" "issn:1825-7321, 1825-7321" ""                          ""                          ""                         
[91] ""                          ""                          ""                          ""                          "" 
```
I will contact CORE to get more info.

pkraker:
Thanks, adding the key to Renviron.site did the trick. This works now as expected - the only thing that needs changing is that `published_in` should not use the ISSN as a fallback as this is more likely to confuse the user than anything else.

pkraker:
One more thing: searchCORE.php still expects a CORE API key to be set in the config.ini. This should be removed as it is not needed anymore with the environment variable approach.

pkraker:
Looks good now, thanks!

PR #270, opened 2018-08-23
fix #261

### LINKED ISSUE #261: VIPER: Maps do not correctly update with the automatization component
In VIPER, maps should update automatically with the automatization component. Currently, however, they do not (always) do that. A quick look at the database shows that some maps are still in revision 1, even though they were created many days ago. I suspect that this may have to do with a change in how IDs for VIPER are created (not using the full set of parameters passed to search.php, but rather only the funder and the project ID).

chreman:
* the subset of params (project_id and funder) which is utilized in searchOpenAIRE is now used as `params_for_id` in the update component as well
* this should solve the issue that maps created under new IDs are not being updated


pkraker:
Looks good, thanks!

PR #271, opened 2018-08-28
Service splitup logging

chreman:
This PR has three main angles - all of them don't change functionality or results:
1) the vis_layout file has been separated into preprocess, cluster, summarize and postprocess, and the functions have been moved. this helps with readability of the code and easier understanding of the workflow. if possible, in each file the functions have been made more readable by splitting them up
2) a logging system has been introduced and all print statements have been replaced by either an info or debug message, additional outputs have been added. this helps with development and also in gaining understanding of how the system performs in production. this change requires the installation of the R package `logging` and the setting of the parameters `OKM_LOGFILE` (e.g. '/path/to/file.log') and `OKM_LOGLEVEL` (e.g. 'INFO') in the Renviron file
3) snapshot tests have been added. they consist of an input-file, simulating results from a base or pubmed-query (`get_papers`), and an output-dataframe, representing the output of a mapping process (`vis_layout`). both files have been produced on the master branch. a snapshot test takes the input-data, executes the mapping (runs vis_layout on the input data), and compares the results with the expected ones.

The snapshot tests for this PR should look like the following, where the only deviations are in the x and y field, where the results are random.
```
Table: Differences detected by variable

var.x            var.y               n   NAs
---------------  ---------------  ----  ----
title            title               0     0
pmid             pmid                0     0
published_in     published_in        0     0
paper_abstract   paper_abstract      0     0
date             date                0     0
year             year                0     0
authors          authors             0     0
subject          subject             0     0
url              url                 0     0
content          content             0     0
doi              doi                 0     0
subject_orig     subject_orig        0     0
readers          readers             0     0
pmcid            pmcid               0     0
x                x                 100     0
y                y                 100     0
area_uri         area_uri            0     0
cluster_labels   cluster_labels      0     0
area             area                0     0
```


pkraker:
Many thanks! I have now conducted a preliminary review. So far, things are looking very good :) But I will need to do some additional tests to confirm the merge. 

Here's a number of smaller issues that I found so far:

* Snapshot test does not work:
```
> summary(compare(output, expected, by="id"))
Error in compare(output, expected, by = "id") : 
  could not find function "compare"
```

* Does not work with DEBUG set to true. But this seems to have been an issue for some time, as it also affects master (see #281).

* Please document setup of logging process (e.g. that the logging file must be created beforehand).

* Prefix R environment variables with HEADSTART rather than OKM

chreman:
1) I did not include the package dependency for the snapshot tests - please install `arsenal` for R (now also added to the server_config.md)
2) env var prefix changed
3) the DEBUG mode currently is broken. the debug functions that export intermediate CSVs and PDF plots rely on previously global variables that can no longer be accessed within the functions where the DEBUG output should happen. We can address this later in a specific debug mode rework to fix #281 but I would suggest to leave it right now, as it is not critical. 
4) Logging documentation added to server_config.md

pkraker:
Thanks! I have now run the snapshot tests, and interestingly enough, there are differences on the Windows machines that I have tested it on (with different R versions). On the Ubuntu server, however, the outcome is as predicted. What would be the best procedure to run the snapshot tests from scratch in Windows?

On a minor note, please also add `knitr` to the list of required packages. In addition, I couldn't find the logging documentation in `server_config.md`. Did you commit the correct file?

pkraker:
Thanks! I also think that encoding may be the culprit here. Could you please add an R file that creates the files necessary to perform a snapshot test (i.e. `snapshot_base_input.json` and `snapshot_base_expected.json`)? Then I can verify that the Windows version gives the same output as before.

server/preprocessing/other-scripts/test/base-snapshot.R
```diff
@@ -0,0 +1,40 @@
...
+options(warn=1)
+
+wd <- dirname(rstudioapi::getActiveDocumentContext()$path)
+
+setwd(wd) #Don't forget to set your working directory
+
+query <- "digital education" #args[2]
+service <- "base"
+params <- NULL
+params_file <- "snapshot_params_base.json"
```
pkraker: I don't think the above paramters are really needed. They are usually passed to get_papers.

server/preprocessing/other-scripts/test/base-snapshot.R
```diff
@@ -0,0 +1,40 @@
...
+
+source('../utils.R')
+source("../vis_layout.R")
+source('../base.R')
```
pkraker: This file does not need to be loaded IMO

chreman:
Here is a small script to create snapshot files - the thing is the snapshots should be created from the master branch, which means that you need to copy over the snapshot_params_base/pubmed.json as well, and after creating the snapshot_inputs and expected they need to be copied over to the rework branch.
```
### FOR BASE

rm(list = ls())

library(rstudioapi)
library(jsonlite)

options(warn=1)

wd <- dirname(rstudioapi::getActiveDocumentContext()$path)

setwd(wd) #Don't forget to set your working directory

query <- "digital education" #args[2]
service <- "base"
params <- NULL
params_file <- "snapshot_params_base.json"

source('../utils.R')
source("../vis_layout.R")
source('../base.R')

DEBUG = FALSE

MAX_CLUSTERS = 15
ADDITIONAL_STOP_WORDS = "english"

if(!is.null(params_file)) {
  params <- fromJSON(params_file)
}

#start.time <- Sys.time()

input_data = get_papers(query, params, limit=120)
write_json(input_data, "snapshots/snapshot_base_input.json")
#end.time <- Sys.time()
#time.taken <- end.time - start.time
#time.taken

output_json = vis_layout(input_data$text, input_data$metadata, max_clusters=MAX_CLUSTERS,
                         add_stop_words=ADDITIONAL_STOP_WORDS, testing=TRUE, list_size=100)

write_json(data.frame(fromJSON(output_json)), "snapshots/snapshot_base_expected.json")


### FOR PUBMED

rm(list = ls())

library(rstudioapi)
library(jsonlite)

options(warn=1)

wd <- dirname(rstudioapi::getActiveDocumentContext()$path)

setwd(wd) #Don't forget to set your working directory

query <- "health education" #args[2]
service <- "pubmed"
params <- NULL
params_file <- "snapshot_params_pubmed.json"

DEBUG = FALSE

source('../utils.R')
source("../vis_layout.R")
source('../pubmed.R')

MAX_CLUSTERS = 15
ADDITIONAL_STOP_WORDS = "english"

if(!is.null(params_file)) {
  params <- fromJSON(params_file)
}

#start.time <- Sys.time()

input_data = get_papers(query, params)
write_json(input_data, "snapshots/snapshot_pubmed_input.json")
#end.time <- Sys.time()
#time.taken <- end.time - start.time
#time.taken

output_json = vis_layout(input_data$text, input_data$metadata, max_clusters=MAX_CLUSTERS,
                         add_stop_words=ADDITIONAL_STOP_WORDS, testing=TRUE)
write_json(data.frame(fromJSON(output_json)), "snapshots/snapshot_pubmed_expected.json")
```

server/preprocessing/other-scripts/test/base-snapshot.R
chreman: Yes, I intentionally left them in to have a reference on what queries the snapshots were created - but we can document that in a different way and remove the parameters.

pkraker:
Thanks! I have now performed additional tests on Windows and Ubuntu. On Ubuntu, everything works fine. On Windows, however, there seems to be a general problem with replicability. I have filed issue #282 for that, which can be addressed later. Merging this now!

PR #273, opened 2018-08-28
Consistent rotation

chreman:
Review/Merge after #272 .

This PR introduces consistent rotation after switching the NMDS routine from ecodist to [metaMDS](http://cc.oulu.fi/~jarioksa/softhelp/vegan/html/metaMDS.html) from the vegan package. metaMDS tries to find a stable solution using several random starts, leading to a similar/identical layout of maps with similar/identical features.


pkraker:
Nice! I tried running `base-test.R`, but it produced the following error:
```
> output_json = vis_layout(input_data$text, input_data$metadata, max_clusters=MAX_CLUSTERS,
+                          add_stop_words=ADDITIONAL_STOP_WORDS, testing=TRUE, list_size=100)
Error in if (msglevel < level) return(invisible(FALSE)) : 
  missing value where TRUE/FALSE needed
```

server/preprocessing/other-scripts/cluster.R
```diff
@@ -0,0 +1,106 @@
...
+
+}
+
+
+get_ndms <- function(distance_matrix, mindim=2, maxdim=2, maxit=500) {
+
+  # Perform non-metric multidimensional scaling
+  # nm <- par.nmds(distance_matrix, mindim=mindim, maxdim=maxdim, maxit=maxit)
+  # nm.nmin = nmds.min(nm)
+  ord <- metaMDS(distance_matrix, k = 2)
```
pkraker: Is that already the parallel version? The manual suggests that MDS could be parallelized as we did with nmds (standard: 3 cores).

chreman:
I added a parallelization parameter to the nmds call.

Please note that I have to do another merge of master into this branch to reconcile the differences in logging behaviour that have piled up along the last three branches after #272 was approved. I apologize for mixing up feature improvements with the introduction of the logging system, but the simultaneous introduction of that vastly sped up the development process.

pkraker:
This branch is now conflicted with master. Please resolve the conflicts before the review.

chreman:
Merged master, only non-rotation related change is a parameter fix for a test which was missing.

pkraker:
This looks excellent, thanks! Merging with no further comments or changes requested.

PR #274, opened 2018-08-28
Core frontend dropdowns

chreman:
* adapted dropdowns to available features
* language option pending implementation


pkraker:
Thanks for this. Here's what needs changing from my perspective: as the maximum resolution for start and end date is year, the options should reflect that. Please use the solution from the DOAJ example.

chreman:
Currently, the official ropensci/rcoreoa package still is without the language param, please for the moment use `devtools::install_github("chreman/rcoreoa")`

pkraker:
Awesome, thanks!

PR #283, opened 2018-09-11
Linkedcat backend

chreman:
* A first draft of the backend, no specific preprocessing introduced
* many fields are defaulting to empty values
* the query is currently a catch-all term for testing purposes only
* test included
* requires setting `LINKEDCAT_SOLR` in the Renvion/Renviron.site with the url of the solr server


chreman:
The testing issues should be solved after #286 has been merged.

pkraker:
Thanks, this does work well now!

PR #286, opened 2018-09-15
logging cleanup

chreman:
This PR cleans up the logging system:
* By default, no environment variables are needed, in this case logging happens only to console (also not std.out), the default loglevel is INFO
* setting up the loggers and the loglevel has been moved to the driving scripts (text-similarity and the test scripts)
* The documentation has been updated accordingly
* the tests for base and pubmed, as well as their snapshot tests run through without unexpected diffs.


pkraker:
Thanks, this looking good!

PR #288, opened 2018-09-18
core with description filter

chreman:
* merge master into core (a lot of stuff happened)
* install new rcoreoa with filter from `remotes::install_github("chreman/rcoreoa@description-filter")`
* run tests and check `sum(is.na(input_data$metadata$paper_abstract))`, should be 0


pkraker:
Thanks!

PR #289, opened 2018-09-19
critical preprocessing bugfix

chreman:
When too many languages are detected previously, the encoder breaks due to a too low threshold.


pkraker:
Great, thanks!

PR #295, opened 2018-09-28
fixed #290

### LINKED ISSUE #290: Error in matrix(nrow = length(res$dcdocid)): data is too long
The backend processing fails in e.g. `edelstahl 1.4303 dcdate:[1665-01-01 TO 2018-09-19] dctypenorm:(121) dcdescription:?` in BASE.
Most probably this is due to not enough papers being found, but as a first step this error needs to be handled better: If this indeed is the case, raise an error at this stage rather fail ungraceful trying to do data processing.

chreman:
For BASE get_papers now stops execution when no results have been returned by the API, and returns an error message to `text_similarity`.


pkraker:
Looks good, thanks!

PR #296, opened 2018-09-28
fixed #291

### LINKED ISSUE #291: Error in deduplicate_titles(metadata, list_size): object input_data not found
This error occurs trying to create the map representation. This is most likely due to a previous error #290 resulting in input_data not being created. If this is the case, the error should be handled better: Log Error as empty input_data, rather than fail ungraceful trying to perform map generation.

chreman:
Now leaves vis_layout early if no input_data exists, returns an error to text_similarity.


pkraker:
Works well, thanks!

PR #297, opened 2018-09-28
fixed #292

### LINKED ISSUE #292: Error in (function (d, method = "complete", members = NULL) : must have n >= 2 objects to cluster
Here, input data is found but the number of papers is too low and clustering fails.
Example query: `chicken eggshell coloration||121 1665-01-01 2018-09-19 most-recent` in BASE results in only two papers.

chreman:
In the rare cases when only 1 paper is returned, clustering fails. This is now handled explicitly and an error returned to text_similarity.
Please merge after #295 and #296 , as this error handling will only become apparent when the previous ones are handled correctly.

### DIFF
--- a/server/preprocessing/other-scripts/cluster.R
+++ b/server/preprocessing/other-scripts/cluster.R
@@ -9,6 +9,9 @@ get_cut_off <- function(css_cluster, attempt=1){
 }
 
 create_clusters <- function(distance_matrix, max_clusters=-1, method="ward.D") {
+  if(nrow(distance_matrix) <= 2){
+    stop("Not enough papers for clustering, N <= 2.")
+  }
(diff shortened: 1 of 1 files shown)


pkraker:
Please provide a test case.

chreman:
Example query: `chicken eggshell coloration||121 1665-01-01 2018-09-19 most-recent` in BASE results in only two papers.

pkraker:
Thanks, but this query returns 2 papers, whereas this solution is for cases where one paper is returned. In addition, I believe it would be better to assign this paper to one cluster by default so that you can actually get a visualization result. It's not going to be a really good map, but any result is better than an error. In addition, it would enable a OKMaps equivalent to [Googlewhacks](https://en.wikipedia.org/wiki/Googlewhack) :)

PR #298, opened 2018-09-28
fixed 293

chreman:
In the rare cases when only one or two papers are present, the NMDS ordination fails. This is now handled by catching this situation, and instead of getting a layout, random X-Y coordinates are assigned to the 1 or 2 papers between [-1, 1].


pkraker:
Please provide a test case.

chreman:
Example query: `edelstahl` on pubmed returns only two papers.

chreman:
I have added a commit that implements you suggestion. We now first attempt an NMDS, if this fails random coordinates are assigned in opposing quadrants of the visualization.

pkraker:
Looking good, thanks!

PR #302, opened 2018-10-05
fix #301

### LINKED ISSUE #301: VIPER: Fails due to invalid crossref response
Example query `Overview of P.S. - Post Scriptum: A Digital Archive of Ordinary Writings (Early Modern Portugal and Spain) (295562)` for `EC` fails at `fill_dois`. The reason is a crossref response with 0 doi candidates.

chreman:
Fixes #301 by checking the response first. If crossref returned an empty response (no candidate DOIs for a searched title), the doi_candidate generation does no longer break. Instead it returns an empty DOI `""`, which was the previous state before the `fill_doi` function.

Example query: `Overview of P.S. - Post Scriptum: A Digital Archive of Ordinary Writings (Early Modern Portugal and Spain) (295562)` with funder `EC` failed at fill_dois.

### DIFF
--- a/server/preprocessing/other-scripts/openaire.R
+++ b/server/preprocessing/other-scripts/openaire.R
@@ -186,7 +186,7 @@ fill_dois <- function(df) {
   }
   if (length(titles) > 1) {
     response <- cr_works(query=queries(titles), async=TRUE)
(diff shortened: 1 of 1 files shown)


pkraker:
Works well, thank you!

PR #304, opened 2018-10-10
Base langsearch

chreman:
* Unified the lang-search APIs so that it works with and without language-parameters
* BASE now takes `lang_id` parameter; if not present, default behavior english is kept
* for other APIs default behavior english is kept


chreman:
Stopwords are automatically taken where they exist in the tm package (see utils.R, get_stopwords), the stopwords we add in resources are additional stopwords that we manually put in. I can check other resources, e.g. from the nltk-library in Python.

PR #305, opened 2018-10-15
Base langsearch

chreman:
* fixed stemming for languages where no stemmer is available, but the language is given by the predefined query parameters


pkraker:
Thanks, this works well now!

PR #306, opened 2018-10-15
Add additional context parameter to map if lang_id is present in params

rbachleitner:
Added the languages array to `default-config.js` since I assume only lang_id comes from the backend, (on context.params).
Languages array could be converted to a dict, would be better than having to a `find` on it.


rbachleitner:
Don't know what to do about localization -  in a way it is localized (to english and native langauge).

pkraker:
Thanks! There are few issues with this PR that need fixing: 
1. Please remove the whitespace before "Language:"
1. Please source languages from `config.options.languages` rather than adding them to `default-config.js` (see `document_types` in `canvas.js` for this approach).
1. When the language is set to `all`, the visualization does not load with the following error:
```
Uncaught TypeError: Cannot read property 'lang_in_lang' of undefined
    at Canvas.drawContext (eval at <anonymous> (headstart.js:150), <anonymous>:560:64)
    at Canvas.drawTitle (eval at <anonymous> (headstart.js:150), <anonymous>:334:18)
    at ModuleManager.call (eval at <anonymous> (headstart.js:84), <anonymous>:65:22)
    at Object.init_start_visualization (eval at <anonymous> (headstart.js:84), <anonymous>:301:26)
    at Channel.publish (eval at <anonymous> (headstart.js:108), <anonymous>:232:23)
    at Mediator.publish (eval at <anonymous> (headstart.js:108), <anonymous>:360:13)
    at MyMediator.publish (eval at <anonymous> (headstart.js:84), <anonymous>:213:45)
    at eval (eval at <anonymous> (headstart.js:60), <anonymous>:197:36)
    at eval (eval at <anonymous> (headstart.js:324), <anonymous>:39:13)
    at eval (eval at <anonymous> (headstart.js:72), <anonymous>:2011:7)
```

rbachleitner:
Done:
- Source `lang` from config.options, check if `lang` found (fixes `lang_id='all'` error)
- Reset `default-config.js` as to not include the languages array
- Hide the following spans
```
#context-dataset_count
#context-paper_count"
#context-funder"
#context-project_runtime
```
when appropriate

rbachleitner:
Also moved all R stuff to the R file and added an apt-get clean to the end.

vis/js/canvas.js
```diff
@@ -539,6 +544,20 @@ class Canvas {
             } else {
                 $("#document_types").hide()
             }
+
+            if (this.paramExists(context.params.lang_id) && this.paramExists(config.options.languages)) {
```
pkraker: This will fail, if config.options does not exist (which it may well do).

rbachleitner:
I know it fails, but what should we do in case there's no config.options.languages? As of know it fails silently and displays no Language:, at least in my tests...

pkraker:
I know, but that was not my point. This will fail with an error when there is no `config.options`.

pkraker:
Perfect, thanks!

pkraker:
Perfect, thanks!

pkraker:
Perfect, thanks!

pkraker:
Perfect, thanks!

pkraker:
Perfect, thanks!

Issue #284, opened 2018-09-12
Enable original case of characters in cluster labels


pkraker:
**Problem description:**
Currently, all candidate labels are transformed to lowercase. After the appropriate labels have been selected, the first character of each label is transformed to uppercase. This works well in a number of scenarios, but has problems in others:

1. Languages that require certain words to be always uppercase (e.g. German). In this case, bigrams and trigrams may be very hard to read
1. In case of acronyms, proper names and words in CamelCase.

**Suggested solution:**
After the labelling procedure, compare the lowercase results to a dict of the candidate labels with original spelling. If there are different uppercase candidates for the same lowercase label, use the candidate that has more occurrences. If they have the same number of occurrences, use the candidate that has more uppercase letters. If both conditions are the same, choose a random candidate.

PR #307, opened 2018-10-18
fixed upper-lowercasing of bubble titles

chreman:
Partially implements #284 but only to the point where the first, most common variant is taken.


pkraker:
This works very well, thanks!

PR #308, opened 2018-10-31
remove deprecated tolower-flag

chreman:
the forgotten commit in addition to #307, removing the deprecated tolower-flag


pkraker:
Perfect, thanks!

PR #315, opened 2018-11-07
Linkedcat

chreman:
* queries now also search in OCR text
* OCR text is added as paper_abstract
* title replacement through upper/lower casing matching improved
* lang_param behavior adapted
* params taken from frontend


pkraker:
Looks good, thanks!

PR #316, opened 2018-11-08
Linkedcat

chreman:
* Pubyear is now searchable via params


chreman:
* Now searches in more fields

chreman:
I added a commit to fix the bug in detecting ocr-text presence.

pkraker:
Looks good, thanks!

PR #317, opened 2018-11-13
removed lang_id from BASE params; robustified lang_id mapping

chreman:
This PR removes the previously introduced but currently unused parameter lang_id from BASE.
* lang_id removed from searchBASE.php -> This for the moment restores compatiliby with legacy map-id generation
* the lang_id validation function in text_similarity and base.R have been robustified. the if-check for parameter presence now reliably returns a boolean value
* fallback is still all/english


pkraker:
Looks good, thanks!

PR #318, opened 2018-11-17
cleaned filenames

chreman:
Filenames for stats-images have been cleaned: `ö` has been replaced with `oe`, ` ` (space) has been replaced with `_`. Files should now be better accessible in web environments.


chreman:
This is done now. Since the files are the same, git/github should recognize it as renaming of files.

pkraker:
Perfect, thanks!

PR #320, opened 2018-11-23
Linkedcat backend

chreman:
Regarding input data:
* the used fields have been adapted to the new SOLR fields
* the text_content used for text_similarity has been reduced to the ocrtext only, since the other fields currently rarely contribute valuable information in that regard

Regarding query functionality:
* the return fields have been specified
* the pub_year filter has been updated to the new SOLR field
* two filter params haven added for inclusion or exclusion of document types (`content_type_a`), queries also work if none of them are specified

Regarding summarization:
* now-defunct conditional lowercasing has been removed as this functionality is superceded by the keyword case-matching

Regarding interaction with external queries:
* params are now expected to include `exclude_content_type` and `include_content_type`, they can also be empty lists (see params_linkedcat.json and searchLinkedCat.php)
* this needs to be reflected in the api-caller before the full pipeline can be tested again

### DIFF
--- a/examples/linkedcat/index.html
+++ b/examples/linkedcat/index.html
@@ -27,6 +27,10 @@
             <input style="height:26px; vertical-align: middle;" type="text" name="q" size="61" required>
             <button type="submit" class="btn" style="background-color: #2856a3; color: white; width:10%; vertical-align: middle;">suchen</button>
             <div id="filter-container"></div>
+            <select name="exclude_content_type[]" multiple="multiple" style="display:none">
+                <option value="Mitgliederverzeichnis" selected>Mitgliederverzeichnis</option>
+                <option value="Bibliografie" selected>Bibliografie</option>
+            </select>
         </form>
     </div>
     <div id="progress" style="max-width:600px; margin: 0px auto;">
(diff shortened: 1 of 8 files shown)


pkraker:
Thanks! What's the reasoning behind `include_content_type`?

chreman:
The idea was to cover the case that users only want to display specific document types, similar as in BASE with article type filters. If none is provided, all are returned.

pkraker:
Ok, thanks!

pkraker:
I added a basic frontend for the new backend params. However, I am getting `status:error`. Can you please check what's going wrong here?

chreman:
Yes, for some reason if nothing is provided for `include_content_type` in this way, R parses it as a character string of length 1 (blank space or something, I don't know who would do that?) instead of e.g. an empty list. This leads to the backend formulating a SOLR query like this `"content_type_a:()"` which then fails on the SOLR server.

Is there a way to explicitly hand over an empty list that R can understand? If not then I'll hardcode this case into the query builder.

pkraker:
I don't think that I can send an empty list via the post data - I'd suggest to send either a list or the string "all".

chreman:
This is a good suggestion. The query builder now expects the default value `all` or a list of length `>= 1`.

pkraker:
Thanks, it works now with the frontend changes. I noticed, however, that `linkedcat-test.R` now fails. After this has been fixed, the branch can be merged.

chreman:
The test parameters have been updated to the new query builder, and the test should now complete.

PR #321, opened 2018-11-25
Error handling

chreman:
So far, when the backend processing fails (either due to input data or a processing problem), no output is returned, and we communicate a general purpose message to the user. This PR implements basic heuristics to detect the most common reasons for failing queries/map generation. This allows to give users better feedback about why queries failed, and allows us to give users suggestions for more successful queries.

Process:
* Instead of returning nothing, the R-backend now runs an error detection heuristic (utils.R, detect_error) and returns a json, for example `{"reason":"probably timeframe too short","message":"No input data found.","status":"error"}`
* The search.php service has been modified to check the output of the call to the R-backend, if the json contains a `status == error`, it will pass this json through, if not, it will continue as usual
* If for some reason the R-backend fails totally (e.g. missing packages or other deployment problems), the current behavior remains
* the POST to searchBASE.php/searchPubmed.php now receives this json as response, and the `reason` field can be used to select and display an appropriate message
* The errors have following hierarchy: short timeframe > query length > typo
* The threshold for query length (4 words) is based on the median length of failing queries (5 words) minus 1, the threshold for time frame length (60 days) is a guesstimate, but covers the option of most recent month

Examples:
* on PubMed: `computer science machine learning 2018-11-25 2018-11-25` results in `{"reason":"probably timeframe too short","message":"No input data found.","status":"error"}`
* on BASE: `natural language processing in indonesian literature 1665-01-01 2017-08-19` results in `{"reason":"probably query length","status":"error"}`
* on BASE: `naturall language processsing 1665-01-01 2018-11-24` results in `{"reason":"probably typo","status":"error"}`

### DIFF
--- a/server/preprocessing/other-scripts/test/base-test.R
+++ b/server/preprocessing/other-scripts/test/base-test.R
@@ -43,11 +43,13 @@ LANGUAGE <- get_api_lang(lang_id, valid_langs, service)
 ADDITIONAL_STOP_WORDS = LANGUAGE$name
 
 #start.time <- Sys.time()
-
+failed <- list()
 tryCatch({
   input_data = get_papers(query, params, limit=120)
 }, error=function(err){
   tslog$error(gsub("\n", " ", paste("Query failed", service, query, paste(params, collapse=" "), err, sep="||")))
+  failed$query <<- query
+  failed$query_reason <<- err$message
 })
 
 #end.time <- Sys.time()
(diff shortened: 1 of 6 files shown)


chreman:
The heuristic is now more flexible, and instead of just one error offers a range of criteria, which may or may not apply simultaneously, e.g. `{"reason":["query length","too specific","timeframe too short"],"status":"error"}` for `natural language processing in indonesian literature 2018-11-25 2018-11-25` or `{"reason":["typo","too specific"],"status":"error"}` for `naturall language processsing 1665-01-01 2018-11-24`

pkraker:
Works very well, thank you!

PR #323, opened 2018-12-13
Linkedcat backend

chreman:
* This PR makes dealing with missing metadata fields more explicit - hopefully this will become unnecessary in the future.
* Additionally, SOLR functionality of extracting context around search keywords is used to generate a number of snippets (max. nr currently 100) from the `ocrtext` field. A snippet for the search "Sitzung" has the form of
```
" Curator der Akademie, Erzherzog Rainer, die feierliche <em>Sitzung</em> am 30. Mai mit einer Ansprache eröffnen"
```
* All snippets of one document are pasted via "\n" right now - any other suggestions/requirements?
* Snippets are available from the metadata under the `snippets` key and would need to replace `paper_abstract` in the visualization.
* The content_type search field IDs in `search_options.js` have been changed to fit the spelling of the corresponding SOLR-field. We match the exact string in `content_type_a_str`, else "bericht" would also retrieve "reisebericht".

Sorry, I expected 7674d1 to have been already merge into linkedcat :/


pkraker:
Thanks!

PR #326, opened 2018-12-17
Linkedcat backend

chreman:
* added dev-flag for highlights
* reduced summarization to title and keyword_a
* fixed german stopword problem


pkraker:
Thanks!

PR #330, opened 2019-01-14
Linkedcat backend

chreman:
I think I finally solved all problems related to highlighting:
* Mismatches in retrieved documents were fixed by including main_title into highlighting as well -> now highlights from the main_title are also in the highlighted text, this captures the case when the search term only appears in main_title
* Other missing highlights were fixed by introducing `hl.maxAnalyzedChars` and increasing its value to 5x the default
* Lastly, parsing errors in the R-solrium library have been fixed. this requires that the package is now installed with `devtools::install_github("chreman/solrium")`. See [fixes](https://github.com/ropensci/solrium/compare/master...chreman:master)
* code cleanup: the hotfix-flag and associated if-else branches have been removed again

The backend has been tested on the full pipeline with "islam" (which previously failed) and "wien"

### DIFF
--- a/doc/server_config.md
+++ b/doc/server_config.md
@@ -44,7 +44,7 @@ Make sure you have the following packages installed:
   * plyr (for better data munging)
   * onehot (for feature engineering)
   * textcat (for language recognition)
-  * solrium (for interfacing with SOLR servers)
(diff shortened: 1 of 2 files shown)


pkraker:
Many thanks, this does indeed solve the problem with highlighting. I checked it with further problematic keywords.

PR #333, opened 2019-02-07
Linkedcat backend

chreman:
This PR fixes bad behaviour in area title generation at the step of stopword-filtering from area titles. The problem occured in cases where summary candidates consist only of single tokens: If all top_n candidates each consist of single tokens, all would get filtered out erroneously, resulting in empty area titles for a paper. This in turn would lead to area title substitution from keywords.
This has been fixed by checking for candidate length > 1 and conditionally apply the filter.
Additionally, keyword separators for linkedcat-metadata have been changed from ',' to ';'.

This PR is opened for review, but since this occurs in a stage of the pipeline that concerns all APIs I'd like to run a few other tests first.

### DIFF
--- a/server/preprocessing/other-scripts/linkedcat.R
+++ b/server/preprocessing/other-scripts/linkedcat.R
@@ -73,6 +73,8 @@ get_papers <- function(query, params, limit=100) {
   metadata$oa_state <- 1
   metadata$subject_orig = metadata$subject
   metadata$relevance = c(nrow(metadata):1)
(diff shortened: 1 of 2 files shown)


pkraker:
I have now tested this with with the Linkedcat example. It does work well, but there is one issue left for Linkedcat: basic classification is still separated by `","` instead of `"; "`. Otherwise, I am awaiting the results of your tests with the other pipelines.

chreman:
I added a commit for the bkl_caption separators and tested it on PubMed and today on BASE after the downtime - no changes are apparent on PubMed, on BASE it also improves the cluster naming.

PR #334, opened 2019-02-11
Linkedcat author suggestions - backend

chreman:
This PR adds a PHP-service that returns a list of data points for authors. It returns a JSON list of lists with following data points: author-GND-id, author name, author document count, author dates:
```
[["(DE-588)11881835X","Pfizmaier, August",186,"1808-1887", ""],
 ["(DE-588)118545426","Hammer-Purgstall, Joseph <<von>>",83,"1774-1856", ""]]
```

The script performs a SOLR facet search to get the unique authors and their document counts, and then iterates over them and queries the additional information - searching for the author_name, and taking the values from the first document returned, the last value is currently a placeholder for an image link. Additional context information can also be added at this stage.

The JSON is stored as a cache file, and in a time span of 24h loaded directly, after which it is refreshed. Suggest functionality is now delayed only by the time of transfering the JSON.

* It requires `linkedcat_suggest_cache` to be added to the config_local.ini as exemplified in the config.ini

### DIFF
--- a/doc/server_config.md
+++ b/doc/server_config.md
@@ -67,6 +67,11 @@ Duplicate config.ini in server/preprocessing/conf/ and rename it to config_local
 * general->services_path: Relative path to the client REST services. Needs to be in the public_html/www directory.
 * calculation->binary: Path to RScript binary
 * connection->sqlite_db: Full path to the sqlite datatabase file. For development purposes, duplicate headstart.sqlite in server/storage/ and rename it to a filename of your choice. Enter the path to this file here. **Make sure that your webserver has write & execute permissions to this file and the containing directory.**
+* connection->linkedcat_solr: URL to the SOLR server, without protocol or additional paths
+* connection->linkedcat_user: user string of your provided authentification credentials
+* connection->linkedcat_pwd: password string of your provided authentification credentials
+* connection->linkedcat_suggest_cache: Absolute path to the cache file for the author suggest functionality, for example "/../server/storage/lc_cache.json". **Make sure that your webserver has write & execute permissions to the containing directory, as it will be created and rewritten by the service.**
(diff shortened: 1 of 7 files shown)


chreman:
I added a few commits that
* removed the facet limit, now all authors are returned
* the sorting of the author_list is now alphabetically
* the config.md has been updated with the new config_ini requirements
* an error handling procedure for the getAuthors() logic has been implemented that attempts to catch any error occurring at the refresh stage, and if this fails, skips the refresh and falls back to the existing cache

pkraker:
Thanks! This works now as expected. I have three remaining comments before merging the PR:
* `max_execution_time` is already addressed [in the server config](https://github.com/OpenKnowledgeMaps/Headstart/blob/master/doc/server_config.md#configuration) and is currently set globally in php.ini rather than in a specific file (suggested setting is 240). In any case, this setting should not be set to 0 for a web application, as this would mean that the web server would execute the script forever, even if it erroneously goes into an endless loop, with potentially high load on memory and CPU.
* The fallback for the cache file works well for the majority of cases, but if an error occurs at the very first execution of the script, a file with an empty array is still being created. This should be prevented.
* When the PHP script fails, it does seem to report an error in either of the Apache logs. This potentially makes it very hard to detect. Suggestion would be to either report an error to the client or to write an error to the error log.

chreman:
I added a few commits that
* removed the dangerous infinite-runtime potential
* reports errors in apache log as well as return JSON (not the same though), but good enough for now I guess
* handles errors in a cold-start situation: if no cache file is present, and creating one files, the script will no longer write an erroneous empty cache and provide updates for 24h

pkraker:
Excellent, thanks!

PR #337, opened 2019-02-21
Linkedcat authorview backend - ID query and additional author query

chreman:
This PR implements
* a switch to non-tokenized author IDs as main search query
* additional retrieval of documents where the searched author is co-author
* the search and the backend now expect e.g. `(DE-588)11881835X` as query

Following changes have been made:
* the faceted search in `getLinkedCatAuthors.php` has been updated to the non-tokenized author ID. the delivered JSON with author suggestions remains structurally unchanged, but now contains only deduplicated author names
* the SOLR query in `linkedcat_authorview.R` has been updated to now also query for the additional author field `author700_0_str`
* since the facet query is now on IDs, author name sorting had to be implemented in PHP
* update of parameters in the search service layer
* update of parameters in the backend-tests

Tested for correct return of author suggestions and map representation. A search based on the ID now returns the unique author name and ID, as well as additional documents where the author is not first author - the count would be 186 instead.
```
> output <- fromJSON(output_json)
> nrow(output)
[1] 191
> unique(output$author100_a)
[1] "Pfizmaier, August" ""                 
> unique(output$author100_0)
[1] "(DE-588)11881835X" ""  
```

### DIFF
--- a/server/preprocessing/other-scripts/linkedcat_authorview.R
+++ b/server/preprocessing/other-scripts/linkedcat_authorview.R
@@ -89,7 +89,7 @@ get_papers <- function(query, params, limit=100) {
 
 build_query <- function(query, params, limit){
   # fields to query in
-  q_fields <- c('author100_0', 'author700_0')
+  q_fields <- c('author100_0_str', 'author700_0_str')
   # fields to return
   r_fields <- c('id', 'idnr',
                 'content_type_a', 'content_type_2',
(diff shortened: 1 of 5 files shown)


pkraker:
Also, I just tested the new author suggestions, and for me, author names are always `null`.

chreman:
Reworked the getAuthorData function to a multi_curl with a sliding window to limit the number of concurrent requests.
Additionally, the parameter reverts have been implemented.

pkraker:
Thanks, this addresses all of the issues raised.

PR #335, opened 2019-02-11
Linkedcat authorview backend

chreman:
This PR implements a new search service for the LinkedCat Authorview. It returns a map representation of possibly more than 100 documents of a single author.

* searchLinkedCatAuthorview.php is added
* linkedcat_authorview.R is added, together with the test and adaptation of text_similarity.R where the document limit is increased

* TODO: proper implementation of parameters in searchLinkedCatAuthorview.php

An open point of discussion is whether this (in connection with #334 ) solves the issue of handing over context parameters.

Note: merge PR pending review of #333

PR #341, opened 2019-02-27
fixed area inconsistencies between bubble-titles and associated docs

chreman:
This PR implements the bugfixes from #335 that have been made to the linkedcat-backend also on the authorview-backend. The code diverged here, as a lot of the mapping functionality is shared, but duplicated in both APIs
A redesign and modularisation of functionalities may be advised.


pkraker:
Thanks! Do you have an author id where that makes a difference for me so that I can review the change?

PR #350, opened 2019-03-26
Linkedcat titleonly

chreman:
This PR introduce API-specific summarization:
* a new parameter "api" gets handed down from text-similarity through vis_layout to the cluster-wise summarization procedure
* at this stage it can be decided which tokens are taken into consideration for bubble-titles
* currently only for the linked-cat API a specific route has been implemented, other APIs remain unchanged
* the tests for base and pubmed have been updated to reflect the change in parameters

A few tests have been conducted on base, pubmed which continue to function unchanged; tests on linkedcat show the intended behavior on both authorview and keyword search. for full functionality on linkedcat #347 should be merged as well.

### DIFF
--- a/server/preprocessing/other-scripts/summarize.R
+++ b/server/preprocessing/other-scripts/summarize.R
@@ -122,8 +122,8 @@ get_cluster_corpus <- function(clusters, metadata, api, stops, taxonomy_separato
       all_subjects = paste(subjects, collapse=" ")
     } else {
       all_subjects = paste(subjects, title_ngrams$bigrams, title_ngrams$trigrams, collapse=" ")
+      all_subjects = gsub(",", ";", all_subjects)
     }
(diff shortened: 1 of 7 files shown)


pkraker:
Thanks, this improves the results in LinkedCat titles. However, it seems that there is an issue with keywords that have a comma in them. For example, searching for "österreich" produces a bubble that reads "Kaiserliche Akademie der Wissenschaften in Wien, Leibniz, Österreich". But the papers in the bubble only contain the keyword "Leibniz, Johann Wilhelm" but not "Leibniz" on its own. In the previous version, the same bubble had the title "Kaiserliche Akademie der Wissenschaften in Wien, Leibniz, Johann Wilhelm".

chreman:
I have added a commit that adresses the issue of wrong keyword separation in the LinkedCat data.

PR #364, opened 2019-06-06
simplified streamgraph x-axis binning

chreman:
With this PR the streamgraph-data has been simplified again, since functionality of binning moved to the frontend.
* there is no longer a separate handling and binning needed for streamgraph-timeseries that are longer than a hardcoded threshold
* unused associated functionality was removed

Tested on maps for Hammer-Purgstall, Calasanza, Brücke, Pfizmaier, and also verified that filling up of missing yearly data still works (see map for Firnhaber).


pkraker:
Thanks, this works very well!

PR #365, opened 2019-06-07
Linkedcat backend metadata updates and bugfixes

chreman:
This PR
* improves keyword-metadata collection by retrieving more keywords from the longer keywordchain `'keyword_a', 'keyword_c', 'keyword_g', 'keyword_t', 'keyword_p', 'keyword_x', 'keyword_z'` (compare maps of Zeissberg, Heinrich)
* "published_in" field is re-implemented to the new SOLR-core through concatenating `search_res$main_title, search_res$host_pubplace, search_res$host_pubyear`
* "oa_state" is no longer generic `1` but depends on the data present in `copyright_until` (see map of Schmidt, Wilhelm)
* Empty bubble titles (1 document bubbles and missing keywords) have been now fixed by falling back to Basisklassifikation, this required a separate, service-specific route (see map for Suchenwirt, Peter). About 4 maps remain where any data fallback fails (Bergman, Johan; Birk, Ernst <<von>>; Haas, Michael; Kandler, Pietro). Behavior of other services (BASE, Pubmed) remains unchanged
* naming convention in the backend changed (api -> service)


pkraker:
Thanks, I only found one issue with the implementation: 
* "published_in" should be a string concatenating `host_maintitle`, `host_pubyear` and `host_pubplace` (in that order).

chreman:
I added a commit to the PR to change that ordering.

pkraker:
Thanks, but please also replace `main_title` with `host_maintitle`. Sorry, should have made that clearer above.

PR #366, opened 2019-06-17
Linkedcat keyword streamgraphs

chreman:
This PR adds functionality to create streamgraphs separately from overview maps.
* In the search options, a visualization type selector is re-implemented
* searchLinkedCat now has a vis_type parameter, taken from search_options and relayed to text_similarity
* text_similarity is modified to parse params first, before setting workflow params, as some workflow params will depend on the selected option
* an additional small bugfix for missing keyword replacement has been smuggled in


pkraker:
Thanks, this works well. Tested with all major integrations.

PR #378, opened 2019-10-10
Linkedcat browseview searchfix

chreman:
This PR should fix the problem that sub-basisklassen are not unique across top_basisklassen (e.g. "Insecta" in Biologie and Geowissenschaften.

To solve this problem, the top_basisklasse is added to all search parameters, from the cached browsetree, over the interface-query builder, the search layer, up to the R-linkedcat query backend.

* SOLR queries for subclasses are now filtered according to their associated top_basisklasse
* maps for unambiguous subclasses now no longer have a duplicatable ID

To see the changes, please compare the behaviour for the combination of e.g. "Biologie - Insecta" and "Geowissenschaften - Insecta" before and after.

It may require to add the top_basisklassen parameter to the map context to help users with disambiguation.
Review/merge after #375


pkraker:
Thanks, well spotted! This now needs to be reflected in the map by adding the top_basisklasse to the title.

PR #387, opened 2019-10-31
crisvis2 data update

chreman:
* changed bubble titles in de/en file
* reassigned question in de/en file

tested locally


pkraker:
Thanks for the quick change!

PR #388, opened 2019-11-06
browseview search fix

chreman:
This PR fixes a query builder bug where queries for sub-bkl failed.


chreman:
Two maps where this can be tested are from top_bkl "Sprach- und Literaturwissenschaft":
* Allgemeine Grammatik
* Arabische Sprache und Literatur

PR #390, opened 2019-11-18
Sec update

chreman:
This PR (courtesy of @tarrow ) closes a range of security issues caused by outdated dependencies/packages, mainly by a multi-version webpack update and a range of smaller package updates.

The jump from webpack 1 to 4 necessitates changes in the module structure: using "headstart" as a global variable became uncommon at webpack 2. To keep functionality the same, there is now a wrapper for the package in place.

Tested as a clean install (no node_modules):
* npm install
* npm run dev/prod
* npm start

Following interactions have been tested on a deployed version:
* Loading a map
* zoom in/zoom out via bubble click
* zoom in via paper title click
* transition between bubbles when zoomed in
* selecting individual paper
* backlink "show all papers in area"
* PDF preview zoomed in from bubble (PDF preview image)
* PDF preview from list
* direct paper outlinks
* Open Access only filter
* list text filter
* list sorting options
* combinations of text filter and options filters


chreman:
Thank you for the comments, I added a few commits to update the documentation regarding version numbers and configs.
Regarding the bug: it is suspected to arise due to changes how Webpack >=2 treats global variables. A possible solution may be in moving the dropdown selectors to the appropriate scope. I suggest to treat this as a bug and handle in a separate PR, as this would require different type of work than it is handled in this PR.

pkraker:
Thanks! This was a big one, but it works very well now 💯

PR #395, opened 2019-11-26
changed english cluster name

chreman:
* Changed english cluster name from Impacts to Implications as requested


chreman:
Commit appended with name changes for the other documents.

PR #406, opened 2020-02-20
Replace phantomjs snapshots with puppeteer

pkraker:
This PR replaces phantomjs, which has been suspended, with puppeteer for map snapshots. This requires a change in the setup in that Node,js 10.17+ is now required.

To move from phantomjs to puppeteer, please carry out the following steps (note: this assumes that snapshots have been working so far):
1. Install Node.js 10.17, if needed
2. Install puppeteer via npm (suggestion would be a global install via `npm i -g puppeteer`)
3. Make changes to your config_local.ini as follows:
3.1 Remove the keys `phantomjs_path` and `snapshot_width` under `[snapshot]`
3.2 Introduce the keys `node_path` (absolute path to your node executable) and `node_modulespath` (absolute path to the node_modules directory where puppeteer is installed) under `[snapshot]`.

Notable changes:
* getChartSVG.js rewritten for puppeteer
* Documentation updated
* Cleanup and improvements of snapshots and the snapshot generation process

Note: this PR has only been tested on Windows so far.


chreman:
Thank you for the update! The implementation works a technical level, I have a few comments / documentation updates that concern deployment:

* In the PR you mentioned `node_modulespath`, while in the actual code and updated documentation it is `nodemodules_path`. I take the second version is correct
* Depending on the operating system, manual installation of additional dependencies for the chromium renderer are required. Known dependencies under Ubuntu 18.04 are `libasound2 libatk1.0 libatk-bridge2.0 libgtk-3-dev`
* While local tests worked fine, server side rendering showed a problem with language characters, see example:
example Under Ubuntu 18.04 the installation of major language character sets for chromium is recommended with `apt-get install -y fonts-ipafont-gothic fonts-wqy-zenhei fonts-thai-tlwg fonts-kacst fonts-freefont-ttf --no-install-recommends`
* This should cover most languages but some tests are recommended.

PR #412, opened 2020-03-13
Enable Boolean queries for BASE

pkraker:
This PR enables all Boolean queries described in the [BASE Interface Guide](https://www.base-search.net/about/download/base_interface.pdf) (see section 4 on page 13) in verbatim mode. The changes have been extensively tested, but further tests would be very useful. Examples include:

* 2019-ncov or sars-cov-2
* (cats and dogs) or (sun and moon)
* research -science
* "knowledge domain visualization" or ("knowledge map" and (research or science))

### DIFF
--- a/examples/search_repos/data-config_base.js
+++ b/examples/search_repos/data-config_base.js
@@ -29,4 +29,6 @@ var data_config = {
     
     embed_modal: true,
     share_modal: false,
+    
+    show_keywords: true,
 };
--- a/server/preprocessing/other-scripts/base.R
+++ b/server/preprocessing/other-scripts/base.R
@@ -39,9 +39,14 @@ get_papers <- function(query, params, limit=100,
   blog$info(paste("Search:", query))
   start.time <- Sys.time()
   
-  # remove unnecessary pluses and minuses
-  query_wt_plus = gsub("((^|\\s))[\\+]+[\\s]*", "\\1", query, perl=T)
-  query_cleaned = gsub("((^|\\s))[\\-]+[\\s]*", "\\1-", query_wt_plus, perl=T)
+  # remove pluses between terms
+  query_wt_plus = gsub("(?!\\B\"[^\"]*)[\\+]+(?![^\"]*\"\\B)", " ", query, perl=T)
+  # remove multiple minuses and spaces after minuses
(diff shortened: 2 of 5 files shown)


chreman:
Thank you! In general this leads to a much better retrieval for complex queries, and allows a better fine tuning of search results.
The query parsing can be tricky, and may require more extensive documentation so that users can get the best out of it.
Examples:
* `2019-ncov or sars-cov-2`: fails
* `"2019-ncov" or "sars-cov-2"`: success
* "2019-ncov" and "sars-cov-2": success
* `"2019-ncov" + "sars-cov-2"`: fails
* `"2019-ncov"+"sars-cov-2"`: success (identical map as with ` and ` but different ID)

Likewise, `(cats+dogs) or (sun+moon)` == `(cats and dogs) or (sun and moon)` but `(cats + dogs) or (sun + moon)` fails.
A quick test showed that wrapping the `+` shorthand with optional space characters (changing the last matching group to `\\s?\\+\\s?`) leads to the same verbatim query.

Additionally, it may make sense to add to the documentation that negative keywords should be added with consecutive `-`, e.g. `science -research -knowledge -theory`, as something like `science -(research or knowledge or theory)` is not correctly parsed 
(-> `textus:science -textus:(textus:research or textus:knowledge or textus:theory)`

pkraker:
Thanks for the thorough review! I updated the regex to cover the problems you described. In addition, I added a preprocessing step that removes unnecessary pluses and minuses. The regex now also deals correctly with multiple parentheses and inverted commas (single and double).

Please check again. There is now a test file in /test that provides a number of test cases.

chreman:
Thank you very much, I think this is a great improvement. I took liberty to add functionality to the test script, it now makes a comparison between actual and expected results and notifies of errors.
Please check that the specifications of expected results in the queries list are correct. 
I already made at one change here for the following test case:
`"Processing fails for a+b . Result: textus:a+textus:b Expected: textus:a textus:b"`

pkraker:
Thanks, I introduced several changes to the preprocessing to accommodate the queries that did not produce the expected result; as a results, pluses are now always replaced by spaces. I also made sure that phrases are not altered in the preprocessing step to enable the user to force searching for the exact term. Please check again!

chreman:
Thank you for the expansion, all tests are now successful. In my view, this is good to go.

PR #414, opened 2020-03-14
Linkedcat backend - Snapshots

chreman:
This PR fixes the last remaining parameter issues and activates snapshots for Linkedcat.

* linkedcat config added to Snapshot backend
* A new route has been added that passes the vis_type parameter on to the snapshot function, with this snapshots are now also possible for streamgraphs.
* A few other files along the pipeline had to be changed as well (fix variable naming inconsistencies), to ensure that variable names that influence both visualization outcome and URL forwarding follow the same convention. Otherwise the Snapshotting backend would not find the visualization in author mode and browse mode.

Functionality and snapshotting has been tested for author overviews and streamgraphs, keyword overviews and streamgraphs, and browseview maps; by manual search awaiting successful loading, and downloading the created PNGs and viewing them.

### DIFF
--- a/server/services/snapshot/headstart_snapshot.php
+++ b/server/services/snapshot/headstart_snapshot.php
@@ -17,27 +17,28 @@
             data_config.show_context = true;
             data_config.create_title_from_context= true;
             data_config.options = options_<?php echo $_GET['service']; ?>.dropdowns;
-            if (<?php echo json_encode($_GET['service']) ?> === "linkedcat" ||
-                <?php echo json_encode($_GET['service']) ?> === "linkedcat_authorview" ||
-                <?php echo json_encode($_GET['service']) ?> === "linkedcat_browseview") {
-                if(<?php echo json_encode(isset($_GET['vis_mode']) && $_GET['vis_mode']) ?> === "authors") {
+            var service = <?php echo isset($_GET['service']) ? json_encode($_GET['service']) : NULL ?>;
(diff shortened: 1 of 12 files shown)


server/services/snapshot/headstart_snapshot.php
```diff
@@ -20,24 +20,24 @@
             if (<?php echo json_encode($_GET['service']) ?> === "linkedcat" ||
                 <?php echo json_encode($_GET['service']) ?> === "linkedcat_authorview" ||
                 <?php echo json_encode($_GET['service']) ?> === "linkedcat_browseview") {
-                if(<?php echo json_encode($_GET['vis_mode']) ?> === "authors") {
+                if(<?php echo json_encode(isset($_GET['vis_mode']) && $_GET['vis_mode']) ?> === "authors") {
```
pkraker: This should rather be something like
`if (isset($_GET['vis_mode'])) echo $_GET['vis_mode']`

Otherwise, it will output `true` or `false` rather than `authors` or `browse`

pkraker: It may be worth doing this once at the beginning of the file and then only echo the variable the parameter was assigned to at this point.

chreman:
Thank you for the review. I made adaptations to search.php to avoid php log warnings that break other service snapshots.
The comparison logic has been adapted to yield the correct result from PHP.
I tried reducing code redundancy in headstart_snapshot.php, but this was not successful and only regressed back to failing snapshots for streamgraphs.

pkraker:
Thanks, this works well now!

Issue #409, opened 2020-03-04
Whitespace at the beginning/end of a query triggers a new map


pkraker:
In today's scavenger hunt at TU Graz, we found that a query with spaces at the beginning or the end triggers a new map, e.g. "sustainable development goals " vs. "sustainable development goals" (entered without inverted commas). This is not intended behaviour, in particular because we trim the input on the server side anyway.

Suggested fix: trim the query in search.php prior to creating the map hash id.

PR #417, opened 2020-03-17
Linkedcat backend - multiword, phrase and negative keyword search

chreman:
This PR updates keyword searches: Multiple words, exact phrases, and negative keywords are now working as intended. Highlighting of more than one word now also works.

Please compare following maps/streamgraphs:
* Kaiser Leopold I
* "Kaiser Leopold I"
* Kaiser Leopold II
* "Kaiser Leopold II"
* Leopold -Kaiser -II
* hund
* hund wolf
* hund -wolf
* -hund wolf
* Ferdinand Wolf
* "Ferdinand Wolf" 
* "Ferdinand Wolf" Houdene
* "Ferdinand Wolf" -Houdene
* Ferdinand Wolf -Mundart
* Ferdinand Wolf Mundart

### DIFF
--- a/server/preprocessing/other-scripts/linkedcat.R
+++ b/server/preprocessing/other-scripts/linkedcat.R
@@ -192,6 +192,10 @@ boost_factors <- list(
   'keyword_label'=30
 )
 
+proximity_factors <- list(
+  'ocrtext'=7
+)
+
 build_authorfield_query <- function(field, query) {
   paste0(field, ':', '"', paste0(gsub("[^a-zA-Z<>]+", "*", query), "*"), '"', add_boost_factor(field))
 }
@@ -203,13 +207,15 @@ build_queryfield_query <- function(field, query) {
     for (i in 1:l_q) {
       query <- gsub('("[\\w+ äöü]+) AND ([äöü\\w \\.]+")', "\\1 \\2", query, perl=TRUE)
     }
-    query <- paste0(field, ':', '(', query, ')', add_boost_factor(field))
+    query <- paste0(field, ':', '(', query, ')', add_boost_factor(field), add_prox_factor(field))
   } else {
-    query <- paste0(field, ':', '"', query, '"', add_boost_factor(field))
+    query <- paste0(field, ':', '"', query, '"', add_boost_factor(field), add_prox_factor(field))
   }
 }
 
-
+add_prox_factor <- function(field) {
+  if (field %in% names(proximity_factors)) paste0('~', proximity_factors[field]) else ""
+}
 
 add_boost_factor <- function(field) {
   if (field %in% names(boost_factors)) paste0('^', boost_factors[field]) else ""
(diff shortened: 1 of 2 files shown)


pkraker:
Thanks! This certainly opens up the number of possible queries for LinkedCat. 

@tanteuschi and I did test larger number of queries, and we found the following issues:
* `die auf die Möglichkeit des Vorhandenseins von` works (both KM and SG, same results)
* `"die auf die Möglichkeit des Vorhandenseins von"`, which should return at least one paper from the query before fails for KM and SG.
* The queries `wien                    akademie   geschichte   ` and `wien                       -akademie              geschichte ` fail.
* The query `historisch-archäologischen` works, but the term is not highlighted in snippets.
* The query `wien akademie geschichte` works well, but when searching within the list for these three terms, only 80 document of the 528 documents are shown.
* Exclusions via `-` work well, but inclusions via `+` do not seem to work (e.g. `wien -akademie geschichte` works, `wien -akademie +geschichte` does not)
* `wien - akademie geschichte` has different results than `wien -akademie geschichte`

chreman:
Thank you for the review and the additional examples. I added a few commits that fix nearly all of them. The only case where I did not achieve improvement is for

> The query wien akademie geschichte works well, but when searching within the list for these three terms, only 80 document of the 528 documents are shown.

I added a proximity factor to the query which in theory should demand that the query terms are no more than 7 tokens away from another. Unfortunately, it was not possible to retrieve highlights in this way where all tokens appear simultaneously. Highlighting of multi-word retrievals works in principal, as the examples of `"die auf die Möglichkeit des Vorhandenseins von"` and `die auf die Möglichkeit des Vorhandenseins von` demonstrate. But for the above case the document retrieval and highlighting produce incoherent results.

pkraker:
Nice, this fixes most of the issues mentions. 

I only found problems for the following two queries:

* ` wien                                  -akademie + geschichte ` fails (note the white space at the beginning, between the words and at the end). Regarding the white spaces at the start and end, please also see #409 - may be worthwhile correcting this already in search.php and fixing another issue alongside this one.
* `wien -"akademie der wissenschaften"` produces a map that includes documents with the phrase "akademie der wissenschaften" (`wien +"akademie der wissenschaften"` and `wien "akademie der wissenschaften"` work well and produce the same output).

We'll move the highlighting issue to a different PR, as this issue also has a frontend component (documents are not correctly counted in the streamgraph it seems).

chreman:
I added a commit that improves handling of negative keyword phrases. Additionally whitespace trimming of queries now happens early on as suggested in #409. This change has been tested on the other integrations as well, with the desired behavior.

pkraker:
Thanks, I tested this, but still had problems with the following queries:

* `wien -"akademie der wissenschaften"` still includes documents, which contain the string "Akademie der Wissenschaften". It's the same btw for wien -"Akademie der Wissenschaften"
* Space before and after queries work well now, but multiple spaces between terms still fail. An example for this is (without inverted commas): "wien      -akademie               geschichte"

chreman:
Thanks. I added a final commit that addresses multiple white spaces early on (search.php), and should improve behaviour on all integrations.
I propose to move the retrieval of documents which still contain excluded terms to another issue for now.

pkraker:
Thanks, this works well now!

PR #418, opened 2020-03-19
Linkedcat highlighting

chreman:
This PR fixes highlighting issues for phrases and negative keywords, merge after #417 

It was tested for a range of single-word, phrases, negative phrases, such as
* Kaiser Ferdinand II
* Kaiser -Ferdinand -II
* griechenland -"spanien" ägypten
* "die auf die Möglichkeit des Vorhandenseins von"
* die auf die Möglichkeit des Vorhandenseins von

Positive queries are highlighted, negative keywords or phrases are not, no regression of wrongly highlighted tokens was found.
In some cases, especially large streamgraphs (e.g. "akademie der wissenschaften"), some papers may exist which will not have any highlight, because the query is found in metadata fields which are not used for highlighting.

### DIFF
--- a/examples/linkedcat/js/data-config_linkedcat.js
+++ b/examples/linkedcat/js/data-config_linkedcat.js
@@ -36,7 +36,7 @@ var data_config = {
     //filter_options: ["all", "open_access"],
     
     show_context_oa_number: false,
-    abstract_large: 1000,
+    abstract_large: 3000,
     
     streamgraph_zoom: false,
     canonical_url: "https://openknowledgemaps.org",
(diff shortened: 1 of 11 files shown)


chreman:
OK, I will have another look, this is what I currently get 
grafik

In regards to the number of snippets, there may have been a misunderstanding: We retrieve more than 1 in the backend (that works), but then restrict it to the first most fitting (according to SOLR ranking) for the delivered metadata field. Of course I can increase the number of delivered snippets.

pkraker:
Thanks, this looks good now!

PR #419, opened 2020-03-23
Enable a fallback AJAX search request on the client

pkraker:
This PR introduces two changes, which enable users of Head Start to implement a fallback AJAX search request, should the original request to the search*.php service not return (this is for example the case with bad connections):

1. A service getLastVersion.php that returns the last version of a map for a given ID (used to check if a map/a specific revision has been created).
2. Add the ability to provide search.php with a precomputed ID (otherwise, the client does not know, what map ID to check for).

This change was tested with all existing examples.


chreman:
Thank you for that! It works as intended. I can also confirm the non-fallback behaviour works as before, tested for the examples.

PR #420, opened 2020-03-24
Fix 197 - reactivate article types option for PubMed

chreman:
This PR reactivates the article-types filter/selector for the PubMed integration.

* Thanks to an update of rentrez, courtesy of @sckott, long parameters (in this case, a list of article types) are now POSTed to PubMed, and no longer trigger a "414 request too large error" on their side. Install from the correct branch.
* The list of article types has been updated to the status in Feb. 2020
* The article types are now a new metadata field "publication_type", "; " separated, e.g. `"Clinical Trial; Journal Article; Multicenter Study; Research Support, Non-U.S. Gov't; Webcast" `
* A backend-test has been implemented that checks the article type of retrieved documents against the query type. In general this works well, although a fallback behavior of the PubMed API can be observed: Should no document of a selected article type could be found, other probable article types are taken into consideration. For "clinical trial", this is e.g. "Clinical Trial, Phase III; Journal Article; Multicenter Study; Randomized Controlled Trial" , for "Guideline" this is e.g. "Practice Guideline", for "Popular Work", this is e.g. "Patient Education Handout".


chreman:
I added a commit that implements a fallback should all article types be deactivated: in that case the query omits the article type filter, as was the case until now.

I tested this for two cases: for your above query, where article types is set via the query itself, it now creates a map identical to the current master when i) all article types are deactivated, ii) all article types are activated.

I also tested this for normal searches, here maps are not identical, but very similar between all/no article types.

pkraker:
Thanks!

Issue #178, opened 2018-02-09
PubMed integration fails on queries with an apostrophe


pkraker:
As reported by @gg927, queries with an apostrophe fail for the PubMed integration. They work fine in BASE.

Steps to reproduce: search for `women's health`

sckott:
in the web interface it works but only gives 5 results https://openknowledgemaps.org/map/b28d514ce3c6a52fa6ecc907098e33cb - I can't replicate the problem on the R side though

chreman:
The original problem can not be reproduced, but another related issue exists:
In our tests the above query is logged as is; but in a server environment it shows up in the logs with an escape character, `women\'s health`.
Any idea where this could occur in R, @sckott ? Or whether we should look to other stages where the query gets handed over.
Locale settings from session.Info():
```
locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C              
 [3] LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8    
 [5] LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8   
 [7] LC_PAPER=en_US.UTF-8       LC_NAME=C                 
 [9] LC_ADDRESS=C               LC_TELEPHONE=C            
[11] LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C
```

sckott:
no idea about the escape character. I'll have a look though

Do you know if the same thing happens with the same search phrase when searching BASE?

sckott:
looking through the code in pubmed.R I can't see a place where that escape character could be inserted.

PR #430, opened 2020-05-05
remove slashes pubmed queries

sckott:
#178


sckott:
the slashes are remove at the top of the `get_papers` fxn, let me know if you'd rather they be removed after the logging step, e.g, but I imagine before the logging step is best since the user didn't put the slashes in

PR #435, opened 2020-06-03
addition of default config for 'backend' parameter for data integrations

chreman:
This PR is an addendum to #433 in which the integration between client and backend was changed through the introduction of the parameter "backend" in headstart-client calls to "getLatestRevision".

While a missing parameter is handled by the backend with a fallback, frontend developers remain uninformed about the existence of this parameter. For this reason, a default for this parameter has been added to the config.js of currently maintained data integrations.

### DIFF
--- a/examples/cris_vis/data-config2_de.js
+++ b/examples/cris_vis/data-config2_de.js
@@ -2,8 +2,6 @@ var data_config = {
     tag: "visualization",
     mode: "local_files",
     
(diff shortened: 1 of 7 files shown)


chreman:
Thanks, I removed the parameter from the CRIS vis example and added it default-config.js.

pkraker:
Thanks, looking good now!

PR #439, opened 2020-06-17
Init frontend docs

tarrow:


chreman:
Thanks for the draft, we will add public documentation in a different way.

PR #443, opened 2020-07-02
fix for doaj source, query needed space between query & "AND"

sckott:
Addresses "DOAJ integration fails" item in trello. There wasn't a space before the second `AND` of the query so that it ended up as `marketing AND bibjson.year:[1809 TO 2017]AND _exists_:bibjson.abstract` which gave no results

Also, deleted a line of code (see line 34) that wasn't supposed to be there AFAICT.


chreman:
Thank you for the bugfix. I updated the associated test and the basic integration works again.

PR #445, opened 2020-07-10
Setup some linting for components

tarrow:
Linting is currently restricted to only modern code
because there are many violations of these linting rules
in the upstream code

The PR introduces the same style of linting
that is used by create-react-app. It appears
to be the industry standard way to lint these kind of apps.

Updated developer documentation to make how to lint clear.


Konstiman:
I tried linting a file with some code-quality error (an unused variable) and it notified me about that with an error. So I suppose it works.

I have a question though: what do you mean by saying the linter does the "code style check"? Is it supposed to check formatting too?

I'm asking because I put all the code in `ExampleModernComponent` into one line and the `npm lint` didn't raise any warnings or errors.

I suggest using [Prettier](https://prettier.io/) for that - it's a tool that works in most of the editors and it makes following the code style rules very easy.

PR #448, opened 2020-07-13
clustering exceptional inputdata hotfix

chreman:
This PR implements a hotfix for rare cases of input data the exhibits features in a way that leads to a breakdown of the interplay between clustering and summarization.
In cases where the input data is mostly self-similar, the clustering algorithm tends toward a cluster size of 1, when instead a higher number is desireable. For these cases, a cluster limit of 15 is now hardcoded, until a proper solution can be devised.


server/preprocessing/other-scripts/cluster.R
```diff
@@ -72,6 +72,12 @@ create_clusters <- function(distance_matrix, max_clusters=-1, method="ward.D") {
 
     groups <- cutree(cluster, k=num_clusters)
 
+    # hotfix for clustering fail
+    if (num_items >= 30 && length(unique(unname(groups))) == 1) {
+      num_clusters <- 15
```
pkraker, line 77: This should probably be `MAX_CLUSTERS` instead of `15`

chreman:
Thanks for the review, MAX_CLUSTERS is now taken from the parameters.

PR #451, opened 2020-07-14
update logging to current standard

chreman:
This PR updates the VIPER example to the current standard by giving it a flexible Matomo site ID with the default value of '1'.

I'm not sure about L207 though, maybe it can be omitted. In any case it also needs to reflect the site_id?


chreman:
Thanks for the review, I removed the noscript tag for now.

PR #455, opened 2020-07-23
Add streamgraph example

Konstiman:
I added a streamgraph example that can be run locally just with `npm start` without deploying anything.

There is a precondition: there must be a `config.js` file in the root directory and its content must be exactly:

```
module.exports = {
    publicPath : "http://localhost:8080/dist/",
    skin : "linkedcat"
};
```

Then the process is the same as with the local files example: start the dev server `npm start`. When the build is done, the streamgraph can be accessed at `http://localhost:8080/examples/local_streamgraph/index.html`.

In future, it might be nice to come up with a solution that doesn't involve config file, but uses for example a system variable or a specific npm run command.


chreman:
Thank you very much, this works as intended.

PR #456, opened 2020-07-24
Use Middleware to trigger mediator actions

tarrow:
Use Redux Middleware to provide a place to transparently
inject a function to be called on zoomOut.

Additionally tweaked the tests and used a mocked store for
snapshots.

n.b. class for streamgraph backlink is not added in this
commit


Konstiman:
Thanks for the PR!

It works well with the normal headstart, there were though some bugs in the streamgraph. I fixed them all.

I also changed things I believe were wrong:

- Initiating store state with `null`, which is a value that doesn't represent a valid state. I replaced it with a valid state value in both subreducers.
- Dispatching the event from backlink: instead of passing `dispatch` into the component, I used the `mapDispatchToProps` function. Based on the tutorials I've seen, this approach is used more often.

Moreover, I returned features that I think are useful:

- The onclick warn saying there was a react component click: Chris wanted this feature for testing.
- The onclick tests.

Then I added some more tests, added the backlink-streamgraph class for the streamgraph version and cleaned things a bit (deleted the obsolete snapshot, removed `hide_backlink` from the code and formatted some files with Prettier).

I'm merging the PR now and I'll continue with changes in my branch.

PR #462, opened 2020-08-06
Refactoring 01 (MVP): heading

Konstiman:
This PR introduces the second part of the MVP refactoring: the refactored heading subcomponent.

The new code can be tested by building the application with `MODERN_FRONTEND=true`.

**Main features:**

Redux:

- The zoom-in action now has a payload: the zoomed-in area (bubble/streamgraph area) name. It's used in the title.
- There is the new `INITIALIZE` action that saves necessary config and context variables into the store. It doesn't copy the whole config and context, it rather selects only the data that is actually needed. The whole process happens in the reducers, so it's testable.
- The localization part of the config is copied though - for now, we decided not to refactor the localization, because it's out of the MVP's scope.
- There is the new `FILE_CLICKED` event used in local files, when a different file from the dropdown is selected.

Components:

- There are two main templates for the heading: `ZoomedInHeading.jsx` and `ZoomedOutHeading.jsx`. The zoomed-out version has a more complicated structure that changes with each setup, so I decided to create a title subcomponent for each major setup. It can be found in the `headingtitles/` folder.
- The logical component, `Heading.js`, chooses the templates for each setup. Then it fills them with the data (title, labels, query...).
- The intermediate layer has a new middleware that triggers a file change in local files when a different file from the dropdown is selected.

Legacy code:

- I had to refactor the `drawTitle()` function in `canvas.js` so I could understand what it does. I split it into multiple helper functions and cleaned it.
- The old code that creates the heading is now executed only if the modern frontend flag is disabled.
- In the mediator, I had to move the mediator call that opens the intro modal if `config.show_intro` flag is enabled. This is because the react elements are rendered a bit later than the legacy code elements.

Tests:

- All new code is tested with a 100% coverage. 

**Other features:**

- fixed `.gitignore` (added `/` to the deployment files paths)

Bugs I already know about:

- In project website, the modal buttons are placed slightly higher. This is caused by different HTML structure - the heading is wrapped into an additional div. I will fix this later when the whole MVP is refactored (then I can get rid of the wrapper div).
- ~~In viper, a space is missing between the title and the "more info" link. I honestly don't know why, because the HTML seems to be the same. I'll have to take a closer look at it.~~ solved
- ~~In linkedcat, a title containing _&lt;&lt;von&gt;&gt;_ is wrongly escaped to _&amp;lt;&amp;lt;von&amp;gt;&amp;gt;_. No idea why, perhaps React does this escaping.~~ solved


vis/js/templates/ZoomedOutHeading.jsx
```diff
@@ -0,0 +1,26 @@
...
+  additionalFeatures,
+}) => {
+  return (
+    // html template starts here
+    <h4>
+      {title}{" "}
+      <a data-toggle="modal" data-type="text" href="#info_modal" id="infolink">
+        <span
+          id="whatsthis"
+          dangerouslySetInnerHTML={{ __html: introIcon }}
```
tarrow, line 16: can we find a way to avoid this? Maybe by making an icon component?

vis/js/templates/headingtitles/BasicTitle.jsx
```diff
@@ -0,0 +1,9 @@
+import React, { Fragment } from "react";
+
+const BasicTitle = ({ title }) => {
+  // dangerously Setting the Inner HTML because of e.g. Covis, where the basic title is HTML
+  // the span must be here so the dangerouslySetInnerHTML works
+  return <span dangerouslySetInnerHTML={{ __html: title }}></span>;
```
tarrow, line 6: similarly here, we probably want to try and separate the html and the messages here. This content should really be a new component/template; this could of course be done in a followup

vis/test/store/initialize.test.js
```diff
@@ -0,0 +1,366 @@
+import { initializeStore } from "../../js/actions";
```
tarrow, line 1: cool; loving having the store tested like this :)

vis/test/store/zoom.test.js
```diff
@@ -32,39 +33,84 @@ describe("zoom state", () => {
...
+        const result = zoomReducer(initialState, zoomOut());
+
+        expect(result).toEqual(expectedResult);
+      });
     });
 
-    it("should handle the internal zoom out", () => {
-      const initialState = true;
-      const expectedResult = false;
+    describe("selectedBubble reducer", () => {
```
tarrow, line 73: If the main code was stuck in separate files this could be too. Also just a style thing but might make the tests more readable (having this in a file called zoom.test.js might make it harder to find)

vis/js/canvas.js
```diff
@@ -328,81 +328,132 @@ class Canvas {
         });
     }
 
-    // Draws the header for this
-    drawTitle(context) {
-        let chart_title = "";
+    /**
+     * Refactored helper function for shortening text that's too long.
+     * @param {String} text 
+     * @param {Number} maxLength
```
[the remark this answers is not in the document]
Konstiman: I don't really insist on using JSDoc, VSCode generated that for me automatically so I filled it out :) Return tags are a good idea, I'll add them.

vis/js/canvas.js
```diff
@@ -328,81 +328,132 @@ class Canvas {
         });
     }
 
-    // Draws the header for this
-    drawTitle(context) {
```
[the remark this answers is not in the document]
Konstiman: Good idea, next time I'll create a separate PR for it.

vis/js/canvas.js
```diff
@@ -328,81 +328,132 @@ class Canvas {
...
-        if (config.show_infolink) {
-            let infolink = ' <a data-toggle="modal" data-type="text" href="#info_modal" id="infolink"></a>';
-            subdiscipline_title_h4.append(infolink);
+        return config.localization[config.language].overview_label;
+    }
+
+    /**
+     * Refactored helper function to understand how the drawTitle main if works.
+     * 
+     * ONLY CALL FROM LEGACY CODE, NEVER WITH modern_frontend_enabled!!!
```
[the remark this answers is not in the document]
Konstiman: You're right, throwing an error is a much better idea than a comment. I'll change it.

vis/js/canvas.js
```diff
@@ -328,81 +328,132 @@ class Canvas {
...
+     */
+    getChartTitle(context) {
+        // This should probably make its way to a more global config
```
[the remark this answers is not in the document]
Konstiman: I have no idea :sweat_smile: This comment was already in the code...

vis/js/templates/ZoomedOutHeading.jsx
Konstiman: Good question... the icon comes from config and it's a fontawesome unicode entity, e.g. `&#xf05a;`. I could create an icon component instead (or even better, directly use fontawesome library for react), but then it wouldn't be possible to change the icon from config so simply - a change in code would be needed.

Let's talk about this at the meeting.

vis/test/store/zoom.test.js
Konstiman: Yeah, I'm actually still not sure how to organize the tests.

My idea is to have a single test file for each action type. This file has two actions though, so it's not a very good example.

An alternative is to have a single test file for each reducer. Is that what you mean? Or is there a better solution?

PR #464, opened 2020-08-18
base work

sckott:
- using latest rbace tag v0.2 on github
- modify base get_papers to include descsize in fields pulled down
- modify base get_papers to include the filter parameter passed down to bs_search
- change base test file to include a descsize filter

use latest rbace tag https://github.com/ropensci/rbace/tree/v0.2.0 `remotes::install_github("ropensci/rbace@v0.2.0")` or `install.packages("rbace", repos = "https://news.ropensci.org")`


chreman:
Additional changes were made so that `min_descsize` is now an external parameter.

PR #469, opened 2020-08-25
Introduce some very lighweight documentation

tarrow:


vis/js/components/README.md
```diff
@@ -0,0 +1,9 @@
+# Components
+React components. See https://reactjs.org/docs/components-and-props.html
+
+Each component should be stored in a separate file.
+
+Each component should have the same name as the file it is contained in.
+
+Each file should export both the plain component and have the redux connected
```
Konstiman, line 8: Is this necessary? Usually if a component is connected to the store, it' used only with the store and never separately. In `Backlink.js`, the plain component is exported too, but that's only because I wasn't sure about how to write tests :grin: I stopped doing this in the next components (`ContextLine.js`, `Heading.js`).

Also, some components don't need Redux at all (`SubdisciplineTitle.js`).

vis/test/component/README.md
```diff
@@ -0,0 +1,2 @@
+# Component tests
+Isolated stand alone tests for components using https://enzymejs.github.io/enzyme/ .
```
Konstiman, line 2: Enzyme turned out to be too weak (no DOM events, deep rendering too complicated etc.), so I started using `react-dom` with `react-dom/test-utils` for rendering and `redux-mock-store` for store mocking.

vis/test/snapshot/README.md
```diff
@@ -0,0 +1,2 @@
+# Snapshot Tests
+Snapshot testing using jest. See https://jestjs.io/docs/en/snapshot-testing for more details
```
Konstiman, line 2: And the components are rendered with `react-test-renderer`.

Konstiman (commented):
I really like your idea to add readme to each folder. That's how we keep documentation close to code.

Generally I agree with the descriptions, they are simple and accurate. I only added some comments to the lines that describe approaches that have changed since the backlink refactoring. Can I change those descriptions myself or do you want to do it?

vis/js/components/README.md
tarrow: Sure; happy to make the change. I was also doing it to make testing easier :)

vis/test/snapshot/README.md
tarrow: :+1:

vis/test/component/README.md
tarrow: :+1:

PR #486, opened 2020-09-17
Flagging from config.js

Konstiman:
This PR introduces the support for using `config.js` to set the modern frontend flag.

An example config:
```
module.exports = {
    publicPath : "http://localhost:8080/dist/",
    skin : "linkedcat",
    modernFrontendEnabled : true
};
```

This way, there's no need to set the path variable, simply setting the flag in `config.js` is enough. The flag value can still be overwritten by the path variable value though.

So for example, building with `modernFrontendEnabled : false` and the path variable `MODERN_FRONTEND=true` will resolve to the flag being set to true.

Special cases:

- Building with `modernFrontendEnabled : true` and the path variable `MODERN_FRONTEND=` will resolve to the flag being set to false.
- Building with `modernFrontendEnabled : true` and the path variable not being set at all will resolve to the flag being set to true.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,8 @@ The tests can be run with:
 
 The modern frontend code can be linted (e.g. the code style etc. checked) using `npm run lint`
 
+Additionally, it is possible to set the flag for the modern frontend in the `config.js` file, for example as `modernFrontendEnabled : true` as demonstrated in `config.examples.js`.
+
 ### Server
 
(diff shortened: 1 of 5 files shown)


chreman:
I also merged a branch that cleans up the hardcoded development environment from webpack, and specifies the environment for the deployment scripts in package.json

PR #489, opened 2020-09-22
changes to base and pubmed get_papers methods to use retry

### LINKED ISSUE #465: RETRY - where in the code do we want to toggle retry options?
rbace and rentrez support RETRY requests now. 

They both allow retry settings to be set by passing in a named list via a fxn, `rbace::bs_retry_options()` and `rentrez::entrez_retry_options()`, to whatever fxn is being used to query the data source. 

Do we want: 

- a new parameter in `get_papers()` for each data source that accepts a named list of retry options that's passed down to the retry settings parameter within the rbace or rentrez fxns
- or just set the retry options internally within each `get_papers()` function so they're hard coded?
- or some other option?

sckott:
- each fxn gains a new parameter `retry_opts`
- the `retry_opts` parameter accepts a call to either the `bs_retry_options` or `entrez_retry_options` function from each respective package
- the retry options is a named list and gets passed to either rentrez or rbace to configure retry http settings
- defaults are used for retry options, but can be set by the user on each `get_papers` function call

fix #465

### DIFF
--- a/server/preprocessing/other-scripts/base.R
+++ b/server/preprocessing/other-scripts/base.R
@@ -11,6 +11,9 @@ library(rbace)
 #    * article_types: in the form of an array of identifiers of article types
 #    * sorting: can be one of "most-relevant" and "most-recent"
 # * limit: number of search results to return
+# * retry_opts: BASE retry options, see `?rbace::bs_retry_options` for documentation.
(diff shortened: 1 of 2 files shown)


sckott:
- rentrez: `remotes::install_github("sckott/rentrez@http-post")`

sckott:
- rbace: `remotes::install_github("ropensci/rbace")`

PR #503, opened 2020-11-05
Postgresql container

chreman:
This PR adds a dockerized Postgres-database as a second option to the backend.

* Updated docker-compose with new services for postgres and pgadmin4
* updated config-files with new `persistence_backend` parameter which defaults to `legacy`, except for triple and gsheets examples
* Created a new persistence-namespace in services/apis, and migrated functionality from sqlite to flask-sqlalchemy
* added tests for persistence layer
* Created alternative routes in search.php, GSheetUpdateAvailable, getGSheetsMap, getLatestRevision, 
* Cleaned up code and added a new function to CommUtils

This PR also contains a new gsheets example, which was developed in parallel and is strongly interdendent with backend-functionalities developed here.


sckott:
woops, sorry, was going to review, but didn't get to it in time.

PR #870, opened 2026-04-13
feat: remove mesh keywords in streamgraph

andreishket:
This PR disables the use of keywords with mesh indexes in the creation of Streamgraph visualisations.


chreman:
This PR now also contains a hotfix that temporarily deactivates logging of browser resize events. These events current are not debounced and tend to create a large burst of logging requests.

